PR #22, opened 2020-11-02
[changelog skip] Install node for apps with execjs

schneems:
The `execjs` gem allows Ruby code to execute javascript. It does this by searching for several possible implementations:

- The ruby racer (V8 distributed as a Ruby gem)
  - This gems is extremely heavy, takes a long time to compile, frequently fails compilation and is generally not commonly used.
- MinRacer (A paired down interface to V8 distrbuted as a Ruby gem)
  - From the readme: Unlike therubyracer, mini_racer only implements a minimal bridge. This reduces the surface area making upgrading v8 much simpler and exhaustive testing simpler.
- System installation of `node`

Prior to the introduction of `mini_racer` the most common and popular development strategy was to depend on a system version of `node`. This PR implements existing behavior in the old buildpack where we install node if `execjs` is present in the gemfile.lock https://github.com/heroku/heroku-buildpack-ruby/blob/38cb9c1e8e16fd09cdf2086387fe9a1882d920de/lib/language_pack/ruby.rb#L1153-L1154. While this might sound like an edge case. Our existing "getting started ruby" example app needs this behavior https://github.com/heroku/ruby-getting-started

Unlike the old Ruby buildpack we do not manually manage and install a node binary, instead we rely on the Node buildpack to install and manage node so that user expectations are more aligned.

## Implementation details

The Node.js buildpack requires a `package.json` copy at the root of the buildpack for detection. In our case the developer may not have added one, but we still need to use the `heroku/nodejs` buildpack to manage our node for us. To accomplish this the Ruby buildpack detects this scenario and writes out a minimal `package.json` so that the nodejs buildpack will not fail detection.

## CNB Concerns

This same strategy does not work for CNB because the detect phase runs in parallel so even if the Ruby buildpack writes a `package.json` to disk as part of it's detect phase, it's too late and the nodejs buildpack's detect has already fired. While this strategy seems to work well enough for the V2 interface, we need to figure out how to support this behavior with the CNB interface.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -6,10 +6,15 @@ BUILD_DIR=$1
 CACHE_DIR=$2
 ENV_DIR=$3
 
-BIN_DIR=$(cd $(dirname $0); pwd)
-BUILDPACK_DIR=$(dirname $BIN_DIR)
+BIN_DIR=$(cd $(dirname "$0"); pwd)
+BUILDPACK_DIR=$(dirname "$BIN_DIR")
 
 source "$BIN_DIR/support/bash_functions.sh"
+
+if needs_package_json "$BUILD_DIR"; then
+  echo "      Writing package.json"
+  echo "{}" > "$BUILD_DIR/package.json"
+fi
 
 if detect_needs_node "$BUILD_DIR"; then
   cat <<EOM
--- a/bin/support/bash_functions.sh
+++ b/bin/support/bash_functions.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 curl_retry_on_18() {
   local ec=18;
   local attempts=0;
@@ -176,19 +178,39 @@ which_node()
   which node
 }
 
-# Returns truthy if the project needs node installed
+# Returns truthy if the project needs node installed but does not
+# have a package.json for example if a Gem in the gemfile depends on node
+needs_package_json()
+{
+  local app_dir=$1
+  local truthy=0
+  local falsey=1
+
+  # If it already has it, don't over-write
+  if [ -f "$app_dir/package.json" ];then
+    return $falsey
+  fi
+
+  if grep -Fq "execjs" "$app_dir/Gemfile.lock";then
+    return $truthy
+  else
+    return $falsey
+  fi
+}
+
 detect_needs_node()
 {
+
+  local app_dir=$1
 
   local needs_node=0
   local skip_node_install=1
 
-  if [ $(which_node) ]; then
+  if which_node; then
     return $skip_node_install
   fi
 
-  local build_dir=$1
-  if [ -f "$build_dir/package.json" ];then
+  if [ -f "$app_dir/package.json" ];then
     return $needs_node
   else
     return $skip_node_install
--- a/hatchet.json
+++ b/hatchet.json
@@ -1,5 +1,6 @@
 {
   "ruby_apps": [
+    "heroku/ruby-getting-started",
     "sharpstone/default_ruby",
     "sharpstone/minimal_webpacker"
   ],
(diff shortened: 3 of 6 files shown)


danielleadams:
> This same strategy does not work for CNB because the detect phase runs in parallel so even if the Ruby buildpack writes a package.json to disk as part of it's detect phase, it's too late and the nodejs buildpack's detect has already fired.

Are you sure this is true? The group order is specified so that each `bin/detect` runs in order: https://github.com/buildpacks/spec/blob/main/buildpack.md#process

Additionally, I need to think about this scenario. Technically, to use Node.js, you don't need a `package.json`, you just need it to use npm or yarn, so maybe we remove that constraint for people that either want to use their own package manager or only need the engine (it is supposed to be the engine buildpack anyways).

hone:
> Are you sure this is true? The group order is specified so that each `bin/detect` runs in order: https://github.com/buildpacks/spec/blob/main/buildpack.md#process

From the spec if you search for "parallel" you come across this line:

> The lifecycle MAY execute each `/bin/detect` within a group in parallel.

> Technically, to use Node.js, you don't need a package.json, you just need it to use npm or yarn, so maybe we remove that constraint for people that either want to use their own package manager or only need the engine (it is supposed to be the engine buildpack anyways).

This is what I was thinking. The node-engine buildpack should always pass and provide `node` but not require it. This way any buildpack can use it. The npm/yarn buildpacks can require `node` and also contain the `package.json` detect logic. The only downside is that by removing the node-engine buildpacks node require, the node-engine buildpack isn't standalone. I'm not sure that is a use case we care about though. With these changes, the nodejs meta buildpack should be the same for end users.

danielleadams:
> The only downside is that by removing the node-engine buildpacks node require, the node-engine buildpack isn't standalone.

Can you elaborate on this? What would standalone entail? I think the only functionality it needs is to install Node (+ npm/npx) and be able to run code with `node file.js`. The requirement for `package.json` would be removed, but the user can still specify a Node version with a `package.json` if needed. @schneems do users specify a version?

hone:
> Can you elaborate on this? What would standalone entail?

Yeah standalone would just be using the node-engine buildpack by itself without another buildpack requiring "node", since my suggestion is for the node-engine buildpack to only provide but not require it. If we just removed detect for `package.json` but kept the provide/requires the same it would always pass, which isn't what Richard wants.

danielleadams:
I see. I have 2 questions then:
- Would it be over engineering to remove the `package.json` detection from the nodejs-engine, then create a buildpack that wraps around nodejs-engine and uses looking for `execjs` to install?
- Is there a way to add an optional buildpack on the fly? Or opting into a group based on the events of `detect`?

schneems:
I think we need something that looks like this:

- node-top-level
  - requires: nodejs if package.json
  - provides: <none>
  - order: [node-always-run]

- node-always-run
  - requires: <none>
  - provides: nodejs
  - order: []

In this scenario the top level node buildpack would be a thin wrapper, all of the logic would be in the "lower" level buildpack.
  
> Would it be over engineering to remove the package.json detection from the nodejs-engine, then create a buildpack that wraps around nodejs-engine and uses looking for execjs to install?

I think that's exactly what we want. 

> Is there a way to add an optional buildpack on the fly? Or opting into a group based on the events of detect?

I think the loose way this is allowed is by having a buildpack that "provides" things, but does not "require" them. But it's not very on-the-fly. I do wish this was a little more explicit, like in this V2 interface I can directly call another buildpack and don't have to make that decision upfront, but I also understand that could get pretty gnarly.

schneems:
This PR as a stand-alone is okay to go. I documented the larger issue here that we can keep open until the CNB interface issue is resolved and we can move forward https://github.com/heroku/heroku-buildpack-ruby-experimental/issues/27

PR #34, opened 2020-11-13
[close #26][changelog skip] Use docker executor on CI

### LINKED ISSUE #26: Switch to CircleCi `setup_remote_docker` instead of using "machine" executor
This might (hopefully) have fewer limitations

```
Thank you for contacting CircleCI Support and thank you for your feedback!
I understand that these limitations can be frustrating and thank you for your understanding.

    Is there a chance y'all could improve the error message here for the failure mode?

I have created an issue in https://github.com/CircleCI-Public/circleci-cli/issues/518 and it will be tracked and prioritized accordingly.

    Is it possible to be able to have access to the docker command from within another docker image? If so then I could switch away from the "machine" executor which seems like it's hitting a lot of edge cases.

Yes, it is possible to use docker commands from docker executor, but you will need to add the setup_remote_docker step into your .circleci/config.yml. You can read more in https://circleci.com/docs/2.0/building-docker-images/

jobs:
  build:
    docker:
      - image: cimg/ruby:2.7.1
    steps:
      - checkout
      - setup_remote_docker:
          version: 19.03.13

```

schneems:
The "machine" executor has a number of limitations and edge cases, not to mention that the RVM installation is very buggy and frequently causes the build to break.

This commit moves to use docker along with setting up "remote docker" https://circleci.com/docs/2.0/building-docker-images/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,22 +55,24 @@ jobs:
       - <<: *hatchet_setup
       - <<: *hatchet
   pack_cnb:
-    machine: true
+    docker:
+      - image: circleci/ruby:2.7
     steps:
       - checkout
+      - setup_remote_docker:
+          version: 19.03.13
       - pack/install-pack
-      - ruby/install:
-          version: "2.7"
       - ruby/install-deps
       - <<: *hatchet_setup
       - <<: *pack_cnb
   docker_commands:
-    machine: true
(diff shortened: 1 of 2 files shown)


schneems:
https://circleci.com/docs/2.0/building-docker-images/#mounting-folders

PR #35, opened 2020-11-23
[changelog skip] CNB Request JVM in Detect

schneems:
This PR makes the "build plan" bash functions more modular by appending to the file instead of overriding it. This allows us to append on jvm without requiring a matrix of feature support.

I've added an integration test for this behavior, but eventually, we need to modify the Ruby builder so that the JVM buildpack is in the default detect order (A similar PR is done in https://github.com/heroku/pack-images/pull/132, though it is pending work described in https://github.com/heroku/heroku-buildpack-ruby-experimental/issues/27).

I'll go ahead and merge this in before the builder work is complete. That work is being tracked in an issue. https://github.com/heroku/heroku-buildpack-ruby-experimental/issues/36

### DIFF
--- a/spec/unit/bash_functions_spec.rb
+++ b/spec/unit/bash_functions_spec.rb
@@ -176,10 +176,32 @@ module HerokuBuildpackRuby
       end
     end
 
-    it "outputs a node+ruby plan when a package.json is present" do
-      Dir.mktmpdir do |dir|
-        build_dir = Pathname(dir)
+    it "outputs to java+ruby when jruby is present" do
+      My::Pathname.mktmpdir do |build_dir|
+        build_dir.join("Gemfile.lock").write <<~EOM
+          RUBY VERSION
+             ruby 2.5.7p001 (jruby 9.2.13.0)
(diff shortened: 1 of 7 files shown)


spec/unit/bash_functions_spec.rb
```diff
@@ -176,10 +174,32 @@ def exec_with_bash_functions(code, stack: "heroku-18")
...
+    it "outputs to java+ruby when jruby is present" do
+      My::Pathname.mktmpdir do |build_dir|
+        build_dir.join("Gemfile.lock").write <<~EOM
+          RUBY VERSION
+             ruby 2.5.7p001 (jruby 9.2.13.0)
+        EOM
+
+        plan_path = build_dir.join("plan.toml")
+        exec_with_bash_functions <<~EOM
+          # Stub out the call to `which node` so we can pretend it does NOT exist on the system
```
Malax, line 186: nit: outdated/copied comment

schneems: Fixed

PR #60, opened 2021-08-06
Enable Dependabot for GitHub Actions [changelog skip]

edmorley:
See:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates#supported-repositories-and-ecosystems

I've also:
* Changed the frequency to daily, so we don't have to wait as long.
* Removed the concurrent PRs limit (since now this repo is under the `heroku` org, the Circle CI concurrency limits are a non-issue).
* Removed the `labels` config which causes the default to be used, which (a) does the same but more (adds lang specific labels), but also auto-creates the labels, which fixed the errors seen in #49 etc.

Refs GUS-W-9678950.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,8 +2,9 @@ version: 2
 updates:
   - package-ecosystem: "bundler"
     directory: "/"
-    open-pull-requests-limit: 1 # Limit concurrent CI runs from executing
     schedule:
-      interval: "weekly"
-    labels:
-      - "dependencies"
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


doriskwan (approved):

schneems:
>      open-pull-requests-limit: 1 # Limit concurrent CI runs from executing

This is for hatchet execution, not any kind of CI limitation. Too many hatchet tests on the same account running in parallel starve each other and then the test suite hits timeout. It's not an issue with this repo yet as there's fewer tests, but it would be in the future.

edmorley:
Oh I didn't realise this repo used Hatchet too, was thinking just Cutlass.

With `daily` the "lots of PRs at once" issue should be less of a problem (so long as loads of PRs don't build up over the weeks, then when dependabot rebases on commit etc).

schneems:
The original design goals/constraint of this re-write was to support v2 and CNB at the same time so all Cutlass tests also have a hatchet equivalent. 

I would also say in the long run unless we can enable massive parallelism in CircleCi execution that we might need to eventually switch cutlass tests to hatchet for test duration performance reasons. Also depending on how closely our implementation of lifecycle on Heroku ends up being to Pack, we might also have an incentive to test on the platform as well.

PR #76, opened 2021-11-11
Switch to BSD 3-Clause License

edmorley:
Since:
- it's the license Salesforce prefers that we use for open source projects:
  https://github.com/salesforce/oss-template/blob/master/license_info.md
- it's what some of our other newer repositories are already using, and it would be great to be consistent
- it's not that different from MIT:
  https://fossa.com/blog/open-source-software-licenses-101-bsd-3-clause-license/

All contributors to this repository so far are Salesforce employees, so there are no external code contributors from whom we need to ask permission before changing the license type.

GUS-W-10140085.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,9 +1,29 @@
-MIT License:
+BSD 3-Clause License
 
-Copyright (C) 2020 Heroku, Inc.
+Copyright (c) 2021, Salesforce.com, Inc.
+All rights reserved.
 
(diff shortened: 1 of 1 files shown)


schneems (approved):

edmorley:
Used skip changelog, because there is not even a `CHANGELOG.md` in this project yet. I think we should not add the check changelog action without first adding a basic changelog.

PR #77, opened 2021-11-11
Remove unused `.env` file

schneems:
This `.env` file is no longer used. It was copied over from the legacy buildpack. In addition we no longer support `cedar-14` or `heroku-16`.

GUS-W-8216066

### DIFF
--- a/buildpack.toml
+++ b/buildpack.toml
@@ -12,17 +12,6 @@ ruby_version = "2.6.6"
   ]
 
   [[publish.Vendor]]
-  # get tokens for logging build tokens
-  url = "https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/ruby.tgz"
-  dir = "."
-  files = ["./.env"]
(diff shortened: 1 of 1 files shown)


schneems:
Failing in part due to https://github.com/buildpacks/pack/issues/1322

edmorley:
> Failing in part due to [buildpacks/pack#1322](https://github.com/buildpacks/pack/issues/1322)

@schneems To be clear, there is still a buildpack bug here, it's just been hidden by that pack bug + a good helping of bad luck until now.

This buildpack needs to either:
1. Populate the vendored Rubies during a packaging step (so that it never tries to download Ruby to the buildpack directory during the build)
2. Continue downloading Ruby during the build, but do so to a temporary directory instead of the buildpack directory.

The buildpack directory is read-only (and has been for years) - buildpacks cannot write there.

edmorley (approved):

PR #79, opened 2021-11-15
Fix pack tests

schneems:
The Ruby buildpack needs a Ruby binary to "bootstrap" its own execution. With v2 this is done at release time and when being used via git dynamically in `bin/compile`.

That same behavior was ported over to this buildpack, however the directory we write to cannot be in the same directory as the buildpack itself. This causes an error:

```
      [INFO] Discovering process types
       [INFO] Procfile declares types -> web
       ERROR: failed to build: exit status 1

       stderr: mkdir: cannot create directory '/cnb/buildpacks/heroku_ruby/0.0.1/vendor/ruby': Permission denied
       ERROR: failed to build: executing lifecycle: failed with status code: 145
```

## Why didn't this break before

Described in buildpacks/pack#1322 the only reason builds were previously working is that another image using the Ruby identifier and the same version were previously installed and they had the files we needed.

## Test Fix

This PR changes moves the behavior of copying binaries out of `bin/compile` for the CNB interface and instead into the `build.sh` interface which the Cutlass testing tool understands https://github.com/heroku/cutlass/.

When cutlass calls `build.sh` it will vendor in the appropriate binaries (if they don't already exist) and then copy everything to the `target` directory.

## Deploy fix

The same process needs to happen before buildpacks are deployed. This buildpack is currently not deployed, only the "legacy" buildpack is github.com/heroku/heroku-buildpack-ruby.

For the v2 deploy process the `publish.Vendor` in the `buildpack.toml` is understood by Heroku's platform/deploy-process.

For CNB, I'm honestly not sure how those files are getting vendored. It actually looks like they're not:

```
$ cd /tmp
$ cat Dockerfile
FROM alpine:3 AS base

FROM public.ecr.aws/heroku-buildpacks/heroku-ruby-buildpack:0.1.3 as buildpack
COPY --from=base / /
```

Then:

```
$ docker build -t inspectbuildpack .
$ docker run -it --rm inspectbuildpack sh
~$ ls cnb/buildpacks/heroku_ruby/0.1.3/vendor/
dotenv.rb     lpxc.rb       syck_hack.rb  toml.rb
```

I'm not sure how this is working unless people are using it without pack and it's manually downloading the binaries every time.

Doing the same trick with `0.1.0` it looks like it's never been vendored:

```
/ # ls cnb/buildpacks/heroku_ruby/0.1.0/vendor/
dotenv.rb     lpxc.rb       syck_hack.rb  toml.rb
```

### DIFF
--- a/package.toml
+++ b/package.toml
@@ -1,0 +1,2 @@
+[buildpack]
+uri = "."
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -23,16 +23,17 @@ def hatchet_path(path = "")
   Pathname(__FILE__).join("../../repos").expand_path.join(path)
 end
 
+RUBY_BUILDPACK = Cutlass::LocalBuildpack.new(directory: Pathname(__dir__).join(".."))
+
 Cutlass.config do |config|
-  # Image pinned to fix CI until https://github.com/heroku/buildpacks-ruby/pull/79 lands. 
-  config.default_builder = "heroku/buildpacks:18@sha256:7590c0bc92e574253e44ef3848579869cfeb610b80de093463ef4a7d8de3ae03"
+  config.default_builder = "heroku/buildpacks:20"
 
   # Where do your test fixtures live?
   config.default_repo_dirs = [hatchet_path("ruby_apps")]
 
   # Where does your buildpack live?
   # Can be a directory or a Cutlass:LocalBuildpack instance
-  config.default_buildpack_paths = [Pathname(__dir__).join("..")]
+  config.default_buildpack_paths = [RUBY_BUILDPACK]
 end
 
 RSpec.configure do |config|
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ references:
   pack_cnb: &pack_cnb
     run:
       name: Run test suite
-      command: PARALLEL_SPLIT_TEST_PROCESSES="4" IS_RUNNING_ON_CI=1 bundle exec parallel_split_test spec/cnb
+      command: bundle exec rspec spec/cnb
   docker_commands: &docker_commands
     run:
       name: Run test suite
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ repos/*
 
 *.gem
 *.gemspec
+vendor/
+target/
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,7 +9,7 @@ GEM
       excon (>= 0.47.0)
       multi_json
     erubis (2.7.0)
-    excon (0.87.0)
+    excon (0.88.0)
     heroics (0.1.2)
       erubis (~> 2.0)
       excon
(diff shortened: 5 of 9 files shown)


edmorley:
I pinned the builder version in #80 to unblock #76, since it's the last PR left for me to close W-10140085.

This PR will need rebasing and can remove the pinned version :-)

edmorley:
> I'm not sure how this is working unless people are using it without pack and it's manually downloading the binaries every time.

I think perhaps the answer is that no-one is using this buildpack at the moment - they are using the previous version of the CNB from `heroku-buildpacks-ruby`, since they will be consuming either from the builder, or from the CNB buildpack registry - both of which have the older CNB project.

schneems:
> I think perhaps the answer is that no-one is using this buildpack at the moment

I know hashicorp's waypoint uses Heroku's ruby buildpack, but I don't know how it's executed.

schneems:
I've worked around the test issue by disabling parallel tests. I updated the quip and also took a stab at a different solution that I believe would allow for parallel test execution via processes https://github.com/heroku/cutlass/pull/38.

It builds off of https://github.com/heroku/cutlass/pull/37 which I think is a good change whether or not we update cutlass to have "exactly once" build.sh semantics.

package.toml
```diff
@@ -0,0 +1,2 @@
+[buildpack]
+uri = "."
```
edmorley, line 2: This is the default value for the file, so this whole file can be removed if preferred.

spec/spec_helper.rb
```diff
@@ -23,16 +23,17 @@ def hatchet_path(path = "")
   Pathname(__FILE__).join("../../repos").expand_path.join(path)
 end
 
+RUBY_BUILDPACK = Cutlass::LocalBuildpack.new(directory: Pathname(__dir__).join(".."))
+
 Cutlass.config do |config|
-  # Image pinned to fix CI until https://github.com/heroku/buildpacks-ruby/pull/79 lands. 
-  config.default_builder = "heroku/buildpacks:18@sha256:7590c0bc92e574253e44ef3848579869cfeb610b80de093463ef4a7d8de3ae03"
+  config.default_builder = "heroku/buildpacks:20"
```
edmorley, line 29: Pleased to see this moving onto Heroku-20, which is the default used by customers.

Since this buildpack didn't previously support Heroku-20 (as it wasn't listed in `buildpacks.toml`), I guess you may also be using Heroku-18 locally?

To change the default locally, run:

```term
pack config default-builder heroku/buildpacks:20
```

edmorley (approved):

package.toml
schneems: This file is what cutlass is using to know whether you've specified a valid dir or not https://github.com/heroku/cutlass/blob/313bf47448011863a77acccb380502d9953c69cb/lib/cutlass/local_buildpack.rb#L74

spec/spec_helper.rb
schneems: Done:

```
$ pack config default-builder heroku/buildpacks:20
Builder heroku/buildpacks:20 is now the default builder
```

PR #93, opened 2022-03-02
Fresh start with Rust

schneems:


src/main.rs
```diff
@@ -0,0 +1,113 @@
+use crate::gemfile_lock::{BundlerVersion, GemfileLock, GemfileLockError, RubyVersion};
```
edmorley, line 1:
```suggestion
#![warn(unused_crate_dependencies)]
#![warn(clippy::pedantic)]
#![allow(clippy::module_name_repetitions)]

use crate::gemfile_lock::{BundlerVersion, GemfileLock, GemfileLockError, RubyVersion};
```

edmorley: Note: You'll also need to work around the false positives from the integration tests, like so:
https://github.com/heroku/procfile-cnb/blob/76e5aa0d5467c46da076ab4ddde6c0b392c3ad20/src/main.rs#L23-L24

Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2021"
+
+[dependencies]
+libcnb = "0.6.0"
+tempfile = "3.3.0"
+flate2 = "1"
+ureq = "2.2.1"
+sha2 = "0.10"
+tar = "0.4"
+toml = "0.5"
```
edmorley, line 13: `toml` is unused

```suggestion
```

Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+serde = "1.0.133"
+url = "2.2.2"
+thiserror = "1.0.30"
+regex = "1.5.4"
+
+[dev-dependencies]
+libcnb-test = { git = "https://github.com/malax/libcnb" }
+ureq = "2.4.0"
```
edmorley, line 21: `ureq` is already specified in `dependencies` so doesn't need to also be in `dev-dependencies`

```suggestion
```

Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "heroku-ruby-buildpack"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+libcnb = "0.6.0"
+tempfile = "3.3.0"
+flate2 = "1"
+ureq = "2.2.1"
```
edmorley, line 10:
```suggestion
ureq = "2.4.0"
```

Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+sha2 = "0.10"
+tar = "0.4"
+toml = "0.5"
+serde = "1.0.133"
```
edmorley, line 14:
```suggestion
serde = "1.0.136"
```

Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "heroku-ruby-buildpack"
+version = "0.1.0"
+edition = "2021"
+
```
edmorley, line 5: If you wanted to specify a minimum supported Rust version, you could add `rust-version` here:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

src/util.rs
```diff
@@ -0,0 +1,95 @@
+use flate2::read::GzDecoder;
+use sha2::Digest;
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::process::{Command, ExitStatus};
+use tar::Archive;
+
+pub fn command_to_str(command: &Command) -> String {
```
edmorley, line 9: Would it be worth having a unit test for this?

tests/integration_test.rs
```diff
@@ -0,0 +1,42 @@
+#![warn(clippy::pedantic)]
+
+use libcnb_test::{BuildpackReference, IntegrationTest};
+use std::io;
+
+#[test]
+#[ignore]
+fn test() {
+    IntegrationTest::new("heroku/buildpacks:20", "tests/fixtures/default_ruby")
+        .buildpacks(vec![BuildpackReference::Crate])
```
edmorley, line 10: This `.buildpacks()` call is equivalent to the default value, so isn't needed:
https://github.com/Malax/libcnb.rs/blob/5b9495d84f605018f9ff61c3b72ca422d975acde/libcnb-test/src/lib.rs#L118

```suggestion
```

tests/integration_test.rs
```diff
@@ -0,0 +1,42 @@
...
+        .run_test(|context| {
+            // On failure, print the stdout
+            println!("{}", context.pack_stdout);
+
+            assert!(context
+                .pack_stdout
+                .contains("---> Download and extracting Ruby"));
+            assert!(context.pack_stdout.contains("Running: bundle install"));
```
edmorley, line 18: `libcnb-test` now comes with `assert_contains!` which provides better UX for this case when the assertion fails (it prints the expected vs actual values rather than just an "expected true, got false").

See:
https://github.com/Malax/libcnb.rs/blob/424cdf94783cd2da10125c1edea540ddcfe6fdc0/examples/execd/tests/integration_test.rs#L21

(there are a few other usages here and below that could use it)

.circleci/config.yml
```diff
@@ -1,91 +1,72 @@
...
-      - <<: *docker_commands
+      - run:
+          name: Install musl
+          command: sudo apt-get update && sudo apt-get install musl-tools --no-install-recommends
+      - run:
+          name: Add musl target
+          command: rustup target add x86_64-unknown-linux-musl
+      - run:
+          name: Run tests
+          command: RUST_BACKTRACE=1 cargo test --all-features --locked -- --include-ignored
```
edmorley, line 65: If you wanted to split this out into separate `unit-tests` and `integration-tests` jobs (where the former can skip the expensive musl/pack/... setup), then the unit tests variant can drop the `--include-ignored`, and the integration test variant instead switch to `-- --ignored`.

See:
https://github.com/heroku/procfile-cnb#test

edmorley (commented):
(all optional :-))

edmorley:
Cancelled Manuel review request to stop the daily reminders in #languages-ops, given the review request is 3 months old.

buildpack.toml
```diff
@@ -1,26 +1,21 @@
...
 [[stacks]]
-id = "heroku-18"
+id = "heroku-20"
 
 [[stacks]]
-id = "heroku-20"
+id = "heroku-22"
+
+[metadata]
+ruby_url = "https://heroku-buildpack-ruby.s3.amazonaws.com/heroku-20/ruby-2.7.4.tgz"
```
edmorley, line 18: I think this `ruby_url` is now unused, since it's generated dynamically elsewhere in the buildpack?

src/layers/ruby.rs
```diff
@@ -0,0 +1,91 @@
...
+            .map_err(RubyBuildpackError::RubyUntarError)?;
+
+        LayerResultBuilder::new(GenericMetadata::default()).build()
+    }
+}
+
+impl RubyLayer {
+    fn download_url(stack: &StackId, version: impl std::fmt::Display) -> Result<Url, UrlError> {
+        let filename = format!("ruby-{}.tgz", version);
+        let base = "https://heroku-buildpack-ruby.s3.amazonaws.com";
```
edmorley, line 66: Using region-specific domain for parity with heroku/heroku-buildpack-ruby/pull/1311

```suggestion
        let base = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com";
```

edmorley, line 66: Using region-specific domain for parity with heroku/heroku-buildpack-ruby/pull/1311

```suggestion
        let base = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com";
```

src/layers/ruby.rs
```diff
@@ -0,0 +1,91 @@
...
+    use libcnb::data::stack_id;
+
+    use super::*;
+
+    #[test]
+    fn test_ruby_url() {
+        let out = RubyLayer::download_url(&stack_id!("heroku-20"), "2.7.4").unwrap();
+        assert_eq!(
+            out.as_ref(),
+            "https://heroku-buildpack-ruby.s3.amazonaws.com/heroku-20/ruby-2.7.4.tgz",
```
edmorley, line 88:
```suggestion
            "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-20/ruby-2.7.4.tgz",
```

edmorley, line 88:
```suggestion
            "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-20/ruby-2.7.4.tgz",
```

buildpack.toml
schneems: Fun story removing these two lines make detect not pass. This might be a bug in libcnb.rs

schneems: The problem is the inconsistency between `RubyBuildpackMetadata` struct when it's removed. But weirdly it doesn't panic or failt to compile it just fails to pass detect ಠ_ಠ.

edmorley: The reason this happens is:
- The `buildpack.toml` file is deserialised by libcnb at runtime (not compile time), and this fails if the toml file doesn't match the provided schema
- This deserialisation in libcnb happens not only in `bin/build` but during `bin/detect` too (since the detect phase may also want metadata from the buildpack.toml, plus libcnb needs to check things like whether the buildpack API version set is a supported one etc)
- Pack CLI doesn't surface detect errors very well (eg you have to pass `--verbose` to get anything useful)

I'd love to improve the detect errors side upstream in Pack/lifecycle/the CNB spec - let's chat about this later in our 1:1 :-)

.circleci/config.yml
```diff
@@ -1,91 +1,72 @@
...
-      command: PARALLEL_SPLIT_TEST_PROCESSES="4" IS_RUNNING_ON_CI=1 bundle exec parallel_split_test spec/unit
-  hatchet_setup: &hatchet_setup
-    run:
-      name: Hatchet setup
-      command: |
-        bundle exec hatchet ci:setup
+executors:
+  rust:
+    machine:
+      image: ubuntu-2004:202201-02
```
edmorley, line 8: There's an Ubuntu 22.04 image out now - can also use the `current` tag to save having to update this all the time :-)

```suggestion
      image: ubuntu-2204:current
```

(https://circleci.com/developer/machine/image/ubuntu-2204)

Cargo.toml
edmorley: If it helps, a quick way to keep these up to date is to use `cargo upgrade` (note: different from `cargo update`), which is part of this plugin:
https://github.com/killercup/cargo-edit#cargo-upgrade

tests/integration_test.rs
```diff
@@ -0,0 +1,41 @@
+#![warn(clippy::pedantic)]
+
+use libcnb_test::{assert_contains, IntegrationTest};
+use std::io;
+
+#[test]
+#[ignore]
+fn test() {
+    IntegrationTest::new("heroku/buildpacks:20", "tests/fixtures/default_ruby").run_test(
```
edmorley, line 9:
```suggestion
    IntegrationTest::new("heroku/builder:22", "tests/fixtures/default_ruby").run_test(
```

buildpack.toml
edmorley: @schneems If I fix the exit code used by libherokubuildpack:
https://github.com/heroku/libherokubuildpack/issues/54

...then I get this error message for such cases (without needing to use `--verbose`):
```
[Error: Internal Buildpack Error]
Cannot read buildpack descriptor (buildpack.toml): TOML deserialization error while reading TOML file: missing field `metadata` at line 26 column 1
```

...which is much better :-)

Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[workspace]
+members = [
+    "buildpacks/ruby",
+    "commons"
```
joshwlewis, line 4: I called this `"common"` in the Node.js buildpack. I'm not sure either is more correct, but maybe we could be consistent?

buildpacks/ruby/src/layers/bundle_install_configure_env_layer.rs
```diff
@@ -0,0 +1,94 @@
...
+## Environment NOT set by this buildpack
+
+The following bundler environment variables are set by another layer:
+
+- `BUNDLE_BIN=<layer-dir>/bin` - Install executables for all gems into specified path.
+- `BUNDLE_PATH=<layer-dir>` - Directs bundler to install gems to this path
+- `GEM_PATH=<layer-dir>` - Tells Ruby where gems are located. Should match BUNDLE_PATH.
+
+*/
+pub struct BundleInstallConfigureEnvLayer;
```
joshwlewis, line 39: This layer seems to apply to most or all bundler commands, not just `bundle install`? I feel like this struct name is a bit overly verbose, and you might be able to get away with `BundlerEnvLayer` without losing context?

buildpacks/ruby/src/layers/bundle_install_create_path_layer.rs
```diff
@@ -0,0 +1,109 @@
...
+- `BUNDLE_BIN=<layer-dir>/bin` - Install executables for all gems into specified path.
+- `BUNDLE_PATH=<layer-dir>` - Directs bundler to install gems to this path
+- `GEM_PATH=<layer-dir>` - Tells Ruby where gems are located. Should match BUNDLE_PATH.
+
+Other environment variables for bundler are configured by another layer that is decoupled
+from dependency storage on disk to miminimize the risk of having to clear dependencies
+to update an environment variable.
+
+*/
+pub struct BundleInstallCreatePathLayer {
```
joshwlewis, line 35: Similar to above, maybe `BundlerPathLayer` or something still gets the point across without being specific to `bundle install` and so long?

buildpacks/ruby/src/layers/bundle_install_configure_env_layer.rs
```diff
@@ -0,0 +1,94 @@
+use crate::RubyBuildpackError;
+use libcnb::data::layer_content_metadata::LayerTypes;
+use libcnb::generic::GenericMetadata;
+use std::path::Path;
+
```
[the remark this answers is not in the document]
schneems: TIL. Thanks!

joshwlewis (approved):
Leaving a thumb to enable you to merge and iterate.

buildpacks/ruby/src/layers/bundle_install_configure_env_layer.rs
```diff
@@ -0,0 +1,94 @@
...
+None
+
+## Set environment variables
+
+- `BUNDLE_CLEAN=1` - After successful `bundle install` bundler will automatically run `bundle clean`.
+- `BUNDLE_DEPLOYMENT=1` - Requires the `Gemfile.lock` to be in sync with the current `Gemfile`.
+- `BUNDLE_GEMFILE=<app-dir>/Gemfile` - Tells bundler where to find the `Gemfile`.
+- `BUNDLE_WITHOUT=development:test:$BUNDLE_WITHOUT` - Do not install `development` or `test` groups via bundle isntall. Additional groups can be specified via user config.
+
+## Environment NOT set by this buildpack
```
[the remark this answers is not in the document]
schneems: I think thats reasonable. I moved some of the comments next to the env var they're modifying in the code. I added a link to the other module instead of re-explaining them. Working on that change now.

buildpacks/ruby/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+mod steps;
+
+#[cfg(test)]
+mod test_helper;
+mod util;
+
+pub struct RubyBuildpack;
+use libcnb::data::build_plan::BuildPlanBuilder;
+
+fn app_needs_java(context: &DetectContext<RubyBuildpack>) -> Result<bool, RubyBuildpackError> {
```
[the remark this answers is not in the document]
schneems: Moving them below the buildpack implementation. I don't love having a `util.rs` that's a holdover from the example Ruby buildpack. I think if code is only used in one file it should probably live in that file.

Cargo.toml
schneems: I used "s" on other directories like `layers` and `steps`. It looks like some of the nodejs buildpack dirs do too.

buildpacks/ruby/src/steps/default_env.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use crate::{
+    layers::{EnvDefaultsSetSecretKeyBaseLayer, EnvDefaultsSetStaticVarsLayer},
+    RubyBuildpack, RubyBuildpackError,
+};
+
+pub struct DefaultEnv;
+
+impl DefaultEnv {
+    pub fn call(
```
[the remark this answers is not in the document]
schneems: Agreed. We talked about it in one meeting and there was one really good idea for naming so that they stood out. But I can't remember what it was.

I went ahead and renamed them. I'm referencing them by full path in my main so they stand out.

buildpacks/ruby/tests/fixtures/ruby-getting-started/README.md
```diff
@@ -0,0 +1,67 @@
+# ruby-getting-started
```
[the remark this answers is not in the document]
schneems: Agreed. I deleted it. I think we should have some capability of pulling in dynamic live github repos like with hatchet, but let's keep that churn out of the git repo.

buildpacks/ruby/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+                .build()
+        };
+        BuildResultBuilder::new()
+            .launch(LaunchBuilder::new().process(default_process).build())
+            .build()
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum RubyBuildpackError {
```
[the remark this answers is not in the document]
schneems: BTW this was a really good comment. I firmed up a lot of the interior visibility. This was especially helpful for as my two broad strokes weakness (that I see) are errors and path-visibility/project-structure.

buildpacks/ruby/src/layers/env_secret_key_base_layer.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl Layer for EnvSecretKeyBaseLayer {
+    type Buildpack = RubyBuildpack;
+    type Metadata = EnvSecretKeyBaseLayerMetadata;
+
+    fn types(&self) -> LayerTypes {
+        LayerTypes {
+            build: true,
+            launch: true,
+            cache: true,
```
edmorley, line 34: I seem to remember (but can't find in the spec right now), that the cache is not guaranteed to be restored in all cases (eg the platform can choose not to, or to expire it). So I think this implementation is at risk of being fragile. It also means we have the same worry about never being able to mass purge caches, as we do for Heroku classic - when it would be great to get away from that situation.

It feels `store.toml` might be the CNB feature intended for long term storage like this?

.github/dependabot.yml
```diff
@@ -1,10 +1,10 @@
 version: 2
 updates:
-  - package-ecosystem: "bundler"
+  - package-ecosystem: "cargo"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
```
edmorley, line 6: Feel free to change these to monthly, if weekly is too often. We've changed a number of other repos to monthly recently. Manual Dependabot runs can be triggered via the GitHub UI now, which combined with monthly can give the best of both worlds.

CODEOWNERS
```diff
@@ -1,2 +0,0 @@
-# Comment line immediately above ownership line is reserved for related gus information. Please be careful while editing.
```
edmorley, line 1: Should a `CODEOWNERS` nested under the `.github` directory, that includes the language team alias be added?

LICENSE
```diff
@@ -1,29 +0,0 @@
-BSD 3-Clause License
```
edmorley, line 1: Did you mean to delete the licence file?

README.md
```diff
@@ -1,179 +1,170 @@
-## Application contract
+## Develoment
```
edmorley, line 1: What do you think about adding an H1 heading to the README with the project name and then a couple of sentence description, to soften the developer-only content here?  (I don't know if you saw the comment about the CNB readmes in https://github.com/heroku/roadmap/issues/19#issuecomment-1227680566)

commons/src/env_command.rs
```diff
@@ -0,0 +1,537 @@
...
+    ) -> Result<EnvCommandResult, EnvCommandError> {
+        (self.on_non_zero_exit)(error).map_err(EnvCommandError::UnexpectedExitStatusError)
+    }
+
+    pub fn to_string<T: Borrow<HashMap<OsString, OsString>>>(
+        command: &Command,
+        show_env_keys: impl Iterator<Item = impl Into<OsString>>,
+        env: T,
+    ) -> String {
+        let escape_pattern = Regex::new(r"([^A-Za-z0-9_\-.,:/@\n])").unwrap(); // https://github.com/jimmycuadra/rust-shellwords/blob/d23b853a850ceec358a4137d5e520b067ddb7abc/src/lib.rs#L23
```
edmorley, line 389: In procfile-cnb, we used `.expect()` for this case rather than `unwrap()`:
https://github.com/heroku/procfile-cnb/blob/1a5e67fdf6ac87532a5c1893a24e123d541ec3a2/src/procfile.rs#L38

commons/src/gem_version.rs
```diff
@@ -0,0 +1,230 @@
...
+            .expect("Internal Error: Invalid Regular Expression!");
+
+            let segment_regex =
+                Regex::new("[0-9]+|[a-z]+").expect("Internal Error: Invalid Regular Expression!");
+
+            if validation_regex.is_match(version_string).unwrap_or(false) {
+                let (segments_l, segments_r) = segment_regex
+                    .find_iter(version_string)
+                    .map(|regex_match| {
+                        let match_string = String::from(regex_match.unwrap().as_str());
```
edmorley, line 64: This is an unwrap in non-test code? Can this ever happen? Should it be an `.expect()` instead to be self-documenting and/or could this be refactored to avoid the unwrap if it's provable not going to occur?

commons/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+pub mod env_command;
```
edmorley, line 1:
```suggestion
#![warn(unused_crate_dependencies)]
#![warn(clippy::pedantic)]

pub mod env_command;
```

buildpacks/ruby/src/test_helper.rs
```diff
@@ -0,0 +1,80 @@
...
+        if !parent.exists() {
+            std::fs::create_dir_all(parent).unwrap();
+        }
+    }
+    std::fs::write(path, "").unwrap();
+    f(path);
+    std::fs::remove_file(path).unwrap();
+}
+
+#[allow(dead_code)]
```
edmorley, line 25: There are 15 usages of `#[allow(dead_code)]` in the repo. I'm presuming some are redundant (and the code is now being used, but the annotation is leftover), and maybe some is genuinely deadcode? It feels like it would be good to remove all of these annotations, along with anything that's actually deadcode :-)

README.md
edmorley, line 1:
```suggestion
## Development
```

README.md
```diff
@@ -1,179 +1,170 @@
...
+  - `SECRET_KEY_BASE=${SECRET_KEY_BASE:-<generate a secret key>}` - In Rails 4.1+ apps a value is needed to generate crypographic tokens used for a variety of things. Notably this value is used in generating user sessions so modifying it between builds will have the effect of logging out all users. Heroku provides a default generated value.
+- Environment variables modified - In addition to the default list this is a list of environment variables that the buildpack modifies:
+  - `BUNDLE_BIN=<bundle-path-dir>/bin` - Install executables for all gems into specified path.
+  - `BUNDLE_CLEAN=1` - After successful `bundle install` bundler will automatically run `bundle clean`.
+  - `BUNDLE_DEPLOYMENT=1` - Requires the `Gemfile.lock` to be in sync with the current `Gemfile`.
+  - `BUNDLE_GEMFILE=<app-dir>/Gemfile` - Tells bundler where to find the `Gemfile`.
+  - `BUNDLE_PATH=<bundle-path-dir>` - Directs bundler to install gems to this path
+  - `BUNDLE_WITHOUT=development:test:$BUNDLE_WITHOUT` - Do not install `development` or `test` groups via bundle isntall. Additional groups can be specified via user config.
```
edmorley, line 60:
```suggestion
  - `BUNDLE_WITHOUT=development:test:$BUNDLE_WITHOUT` - Do not install `development` or `test` groups via bundle install. Additional groups can be specified via user config.
```

README.md
```diff
@@ -1,179 +1,170 @@
...
+  - `JRUBY_OPTS="-Xcompile.invokedynamic=false"` - Invoke dynamic is a feature of the JVM intended to enhance support for dynamicaly typed languages (such as Ruby). This caused issues with Physion Passenger 4.0.16 and was disabled [details](https://github.com/heroku/heroku-buildpack-ruby/issues/145).
+  - `RACK_ENV=${RACK_ENV:-"production"}` - An environment variable that may affect the behavior of Rack based webservers and webapps.
+  - `RAILS_ENV=${RAILS_ENV:-"production"}` - A value used by all Rails apps. By default Rails ships with three environments: `development`, `test,` and `production`. We recommend all apps being deployed to use `production` and recommend against using a custom env such as `staging` [details](https://devcenter.heroku.com/articles/deploying-to-a-custom-rails-environment).
+  - `SECRET_KEY_BASE=${SECRET_KEY_BASE:-<generate a secret key>}` - In Rails 4.1+ apps a value is needed to generate crypographic tokens used for a variety of things. Notably this value is used in generating user sessions so modifying it between builds will have the effect of logging out all users. Heroku provides a default generated value.
```
edmorley, line 53:
```suggestion
  - `SECRET_KEY_BASE=${SECRET_KEY_BASE:-<generate a secret key>}` - In Rails 4.1+ apps a value is needed to generate cryptographic tokens used for a variety of things. Notably this value is used in generating user sessions so modifying it between builds will have the effect of logging out all users. Heroku provides a default generated value.
```

README.md
```diff
@@ -1,179 +1,170 @@
...
+      - ".sprockets-manifest-*.json"
+      - "manifest-*.json"
+    - We will abort your build if the `rake assets:precompile` task fails
+    - We will run `rake assets:clean` on your app. (Reference: https://github.com/heroku/buildpacks-ruby/blob/d526a49f81becaf571329a1adf5fff0668a6b99a/lib/heroku_buildpack_ruby/assets_precompile.rb#L67-L73, reference: https://github.com/heroku/heroku-buildpack-ruby/blob/951fc728979695990c32df2d4d60ae2d6f6f61c2/lib/language_pack/rails4.rb#L83-L94)
+      - We will cache the contents of `public/assets` if `assets:clean` exists on your application. (pending https://github.com/buildpacks/spec/blob/main/buildpack.md#slice-layers support in libcnbrs)
+      - We will cache asset "fragments" directories if the `assets:clean` exists on the system. (pending https://github.com/buildpacks/spec/blob/main/buildpack.md#slice-layers support in libcnbrs)
+        - We will limit or prune the size of the asset cache in `tmp/cache/assets` to 100 MiB.
+- Process types
+  - Given an application with the `railties` gem we will run `bin/rails server` while specfying `-p $PORT` and `-e $RAILS_ENV"` by default as the `web` process. Use the `Procfile` to override.
```
edmorley, line 47:
```suggestion
  - Given an application with the `railties` gem we will run `bin/rails server` while specifying `-p $PORT` and `-e $RAILS_ENV"` by default as the `web` process. Use the `Procfile` to override.
```

README.md
```diff
@@ -1,179 +1,170 @@
...
+  - We will install gem dependencies using `bundle install`
+  - We will run `bundle clean` after a successful `bundle install` via setting `BUNDLE_CLEAN=1` environment variable.
+  - We will cache the contents of your gem dependencies.
+      - We will invalidate the dependency cache if your stack changes.
+      - We will invalidate the dependency cache if your Ruby version changes.
+      - We may invalidate the dependency cache if there was a bug in a prior buildpack version that needs to be fixed.
+- Gem specific behavior - We will parse your `Gemfile.lock` to determine what dependencies your app need for use in specializing your install behavior (i.e. Rails 5 versus Rails 4). The inclusion of these gems may trigger different behavior:
+  - `railties`
+  - [TODO] List is incomplete
+- Applications without `rake` in the `Gemfile.lock` or a `Rakefile` varient MAY skip rake task detection.
```
edmorley, line 34:
```suggestion
- Applications without `rake` in the `Gemfile.lock` or a `Rakefile` variant MAY skip rake task detection.
```

buildpacks/ruby/src/steps/rake_assets_precompile.rs
```diff
@@ -0,0 +1,295 @@
...
+                HasRakefile(false) => CanRunRake::MissingRakefile,
+            },
+            HasGem(false) => CanRunRake::NoRakeGem,
+        }
+    } else {
+        CanRunRake::AssetManifestSkip(asset_manifests.0.clone())
+    }
+}
+
+/// Checks directory for rakefile varients
```
edmorley, line 141:
```suggestion
/// Checks directory for rakefile variants
```

README.md
```diff
@@ -1,179 +1,170 @@
...
-### attr_reader for values that may be falsey
+### Known differences against `heroku/heroku-buildpack-ruby`
 
-When working with a variable that may be falsey, prefer to use an `attr_reader` over accessing the instance variable directly. This will guard against spelling mistakes. For example:
+This buildpack does not port all behaviors of the original buildpack for Ruby (https://github.com/heroku/heroku-buildpack-ruby). This buildpack is also known as `v2` as it implements version 2 of the heroku buildpack interface (instead of the Cloud Native Buildpack interface).
 
-```ruby
-class Foo
-  private; attr_reader :value; public
+- Rails 5+ support only. The v2 buildpack supports Rails 2+. There are significant maintenace gains for buildpack authors [starting in Rails 5](https://blog.heroku.com/container_ready_rails_5) which was released in 2016. In an effort to reduce overall internal complexity this buildpack does not explicitly support Rails before version 5.
```
edmorley, line 103:
```suggestion
- Rails 5+ support only. The v2 buildpack supports Rails 2+. There are significant maintenance gains for buildpack authors [starting in Rails 5](https://blog.heroku.com/container_ready_rails_5) which was released in 2016. In an effort to reduce overall internal complexity this buildpack does not explicitly support Rails before version 5.
```

README.md
```diff
@@ -1,179 +1,170 @@
...
+- Rails support is now based on the above application contract. Previously there was no official policy for dropping support for older Rails versions. Support was based on whether or not an older version could run on a currently supported Ruby version. With [Rails LTS](https://railslts.com/) this can mean a very old version of Rails. Now we will actively support Rails versions currently [under the Rails core maintenance policy](https://guides.rubyonrails.org/maintenance_policy.html) provided those versions can run on a supported Ruby version. As Rails versions go out of official maintenance compatability features may be removed with no supported replacement.
```
edmorley, line 104:
```suggestion
- Rails support is now based on the above application contract. Previously there was no official policy for dropping support for older Rails versions. Support was based on whether or not an older version could run on a currently supported Ruby version. With [Rails LTS](https://railslts.com/) this can mean a very old version of Rails. Now we will actively support Rails versions currently [under the Rails core maintenance policy](https://guides.rubyonrails.org/maintenance_policy.html) provided those versions can run on a supported Ruby version. As Rails versions go out of official maintenance compatibility features may be removed with no supported replacement.
```

README.md
```diff
@@ -1,179 +1,170 @@
...
+- Caching of `public/assets` is gated on the presence of `rake assets:clean`. Previously this behavior was gated on the existance of a certain version of the Rails framework.
+- Caching of `tmp/cache/assets` (fragments) is gated on the presence of `rake assets:clean`. Previously this behavior was gated on the existance of a certain version of the Rails framework.
```
edmorley, line 110:
```suggestion
- Caching of `public/assets` is gated on the presence of `rake assets:clean`. Previously this behavior was gated on the existence of a certain version of the Rails framework.
- Caching of `tmp/cache/assets` (fragments) is gated on the presence of `rake assets:clean`. Previously this behavior was gated on the existence of a certain version of the Rails framework.
```

README.md
```diff
@@ -1,179 +1,170 @@
...
+      - We will cache asset "fragments" directories if the `assets:clean` exists on the system. (pending https://github.com/buildpacks/spec/blob/main/buildpack.md#slice-layers support in libcnbrs)
+        - We will limit or prune the size of the asset cache in `tmp/cache/assets` to 100 MiB.
+- Process types
+  - Given an application with the `railties` gem we will run `bin/rails server` while specfying `-p $PORT` and `-e $RAILS_ENV"` by default as the `web` process. Use the `Procfile` to override.
+  - If the `railties` gem is not present we will run `rackup` while specifying `-p $PORT` and `-h 0.0.0.0` by default as the `web` process. Use the `Procfile` to override. This requires the `rack` gem and a `config.ru` file.
+- Environment variable defaults - We will set a default for the following environment variables:
+  - `JRUBY_OPTS="-Xcompile.invokedynamic=false"` - Invoke dynamic is a feature of the JVM intended to enhance support for dynamicaly typed languages (such as Ruby). This caused issues with Physion Passenger 4.0.16 and was disabled [details](https://github.com/heroku/heroku-buildpack-ruby/issues/145).
```
edmorley, line 50:
```suggestion
  - `JRUBY_OPTS="-Xcompile.invokedynamic=false"` - Invoke dynamic is a feature of the JVM intended to enhance support for dynamically typed languages (such as Ruby). This caused issues with Physion Passenger 4.0.16 and was disabled [details](https://github.com/heroku/heroku-buildpack-ruby/issues/145).
```

commons/src/env_command.rs
```diff
@@ -0,0 +1,537 @@
...
+    ) -> &mut Self {
+        self.show_env_keys = keys
+            .into_iter()
+            .map(std::convert::Into::into)
+            .collect::<Vec<OsString>>();
+        self
+    }
+
+    // Command is not clonable because it can contain things that are not
+    // clonable such as file descriptors. Instead we remember the
```
edmorley, line 205:
```suggestion
    // Command is not cloneable because it can contain things that are not
    // cloneable such as file descriptors. Instead we remember the
```

commons/src/in_app_dir_cache_layer.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// # Caches a folder in the application directory
+///
+/// Layers are used for caching, however layers cannot be inside of the app directory.
+/// This layer can be used to hold a directory's contents so they are preserved
+/// between deploys.
+///
+/// The primary usecase of this is for caching assets. After `rake assets:precompile` runs
+/// file in `<app-dir>/public/assets` need to be preserved between deploys. This allows
+/// for faster deploys, and also allows for prior generated asssets to remain on the system
```
edmorley, line 18:
```suggestion
/// for faster deploys, and also allows for prior generated assets to remain on the system
```

buildpacks/ruby/src/layers/env_secret_key_base_layer.rs
```diff
@@ -0,0 +1,67 @@
...
+use libcnb::layer_env::{LayerEnv, ModificationBehavior, Scope};
+use rand::Rng;
+use serde::{Deserialize, Serialize};
+use std::path::Path;
+
+pub(crate) struct EnvSecretKeyBaseLayer;
+
+/// # Set the `SECRET_KEY_BASE` environment variable
+///
+/// This environment variable is primarially used for encrypting/decrypting user sessions.
```
edmorley, line 15:
```suggestion
/// This environment variable is primarily used for encrypting/decrypting user sessions.
```

buildpacks/ruby/src/layers/bundle_path_layer.rs
```diff
@@ -0,0 +1,97 @@
...
+///
+/// Must run before `execute_bundle_install_layer` to create directory for dependencies.
+///
+/// ## Layer dir: Create directory for dependencies
+///
+/// Dependencies installed via `bundle install` will be stored in this layer's directory.
+/// This is accomplished via configuring bundler via environment variables
+///
+/// Other environment variables for bundler are configured by another layer that is decoupled
+/// from dependency storage on disk to miminimize the risk of having to clear dependencies
```
edmorley, line 21:
```suggestion
/// from dependency storage on disk to minimize the risk of having to clear dependencies
```

edmorley (commented):
Great work! :-)

(Sorry for all the spelling review suggestions, I opened the PR in my IDE to ease review and I have a spell-checker VSCode extension so they all jumped out)

buildpacks/ruby/src/layers/env_secret_key_base_layer.rs
Malax: I'm not familiar with `SECRET_KEY_BASE`, but I assume this is used to encrypt sessions in cookies to they can't be tampered with on the client side? If so, if the cache would be dropped (like @edmorley mentioned) all user sessions will be invalidated. 

`store.toml` is meant as a persistent metadata store and probably, as @edmorley suggested, the better choice here. I have unresolved this conversation as I think we should at least have some sort of decision record what to do here - even if that decision is "we will still do it as a layer".

schneems: I agree it should not be in a cache. This data needs to persist beyond many deploys and would force log everyone out if lost. I found store.toml related code here https://github.com/heroku/libcnb.rs/blob/b494b0baeef39015be7b5773402e903dfc9b3919/libcnb/src/runtime.rs#L197 so I see I can set it via a builder, but I'm not sure where I can read that data back in via a libcnb interface.

LICENSE
schneems: No, good catch. Thanks, updated.

CODEOWNERS
schneems: Fixed

buildpacks/ruby/src/layers/env_secret_key_base_layer.rs
schneems: Updated to use `store.toml` instead of the cache. Take a look. https://github.com/heroku/libcnb.rs/pull/531 would clean up the implementation a bit.

commons/src/env_command.rs
edmorley: This review comment isn't resolved

commons/src/gem_version.rs
edmorley: This review comment isn't resolved

commons/src/lib.rs
edmorley: This review comment isn't resolved

buildpacks/ruby/src/test_helper.rs
edmorley: This review comment isn't resolved

buildpacks/ruby/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "heroku-ruby-buildpack"
+# This crate is not published, so the only version that is used is the one in buildpack.toml.
+version = "0.0.0"
+publish = false
+edition = "2021"
+rust-version = "1.64"
```
edmorley, line 7:
```suggestion
rust-version = "1.66"
```

buildpacks/ruby/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+libcnb = "0.11.4"
+flate2 = "1"
+ureq = "2.5.0"
+tar = "0.4"
+serde = "1.0.145"
+url = "2.3.1"
+thiserror = "1.0.36"
+regex = "1.6.0"
+rand = "0.8.5"
+glob = { git = "https://github.com/rust-lang/glob" } # https://github.com/rust-lang/glob/issues/117
```
edmorley, line 21: It looks like a new version of the `glob` crate is out, so this can now be changed to use a released version rather than a Git URL:
https://github.com/rust-lang/glob/issues/117#issuecomment-1373720674

buildpacks/ruby/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+tempfile = "3.3.0"
+
+[dev-dependencies]
+libcnb-test = ">= 0.11.4"
```
edmorley, line 25: I wouldn't use a greater than range here, since it will quietly pull in breaking changes with a `Cargo.lock` refresh, whereas the version format below is equivalent to `^0.11.4` in node semver, which will pick up non-breaking changes via `Cargo.lock` refreshes, but require a Dependabot to update `Cargo.toml` to pick up the breaking change - which is what we want.

```suggestion
libcnb-test = "0.11.4"
```

edmorley:
Does this need a `CHANGELOG.md` entry? Currently the skip changelog label has been applied.

buildpacks/ruby/Cargo.toml
schneems: My thinking here was that `libcnb` depends on libcnb-data which comes from `libcnb`. I don't want to accidentally run different versions of `libcnb` and `libcnb-test`. I'm a little fuzzy on the resolution rules it seems like it's not like ruby in that it allows for different versions of the same library as sub-dependencies, but according to https://stackoverflow.com/a/51722134/147390 

> cargo will attempt to find a common version which satisfies all requirements

My hope with this line is that cargo is smart enough to realize we've declared we want `libcnb` 0.11.4 which has a hard dependency on `libcnb-data` 0.11.4 and that asking for `libcnb-test` >= X will resolve to match the same version as libcnb due to the shared mutual dependency of `libcnb-data`.

Otherwise if we forget and only update one but on the other, you can run into weird type errors where it will say things like "expecting struct Store to be Store" just because they're different versions. 

I could be under/over thinking this though. I would like to hear your thoughts.

buildpacks/ruby/Cargo.toml
schneems: Updated in the latest commit

commons/src/lib.rs
schneems: This is such a great catch. I must have resolved by accident. This generates a TON of changes needed. Also it's clear work needs to be done in commons. Thanks a ton for the review.

schneems: Updated, thanks again.

buildpacks/ruby/Cargo.toml
schneems: I updated it with your suggestion. Still curious on your thoughts above though

buildpacks/ruby/src/test_helper.rs
schneems: I've removed all the dead_code I could find

commons/src/gem_version.rs
schneems: Updated to remove the unwrap. This is only needed for `fancy_regex` however we have both that and `regex` loaded, so I switched to using it here

schneems:
> Does this need a CHANGELOG.md entry? Currently the skip changelog label has been applied.

I added a CHANGELOG

edmorley (commented):
Thank you for the changes - looks good, just spotted a couple of small final things :-)

commons/src/gem_list.rs
```diff
@@ -0,0 +1,164 @@
...
+                }, //
+            )
+            .collect::<HashMap<String, GemVersion>>();
+
+        Ok(GemList { gems })
+    }
+}
+
+#[cfg(test)]
+
```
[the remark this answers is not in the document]
schneems: Think the problem is the extra newline? Github won't let me apply your commit as it says there's nothing changed. Updating manually.

buildpacks/ruby/tests/integration_test.rs
```diff
@@ -0,0 +1,178 @@
...
+  remote: https://rubygems.org/
+  specs:
+PLATFORMS
+  java
+RUBY VERSION
+   ruby 2.6.8p001 (jruby 9.3.6.0)
+DEPENDENCIES
+"#,
+    )
+    .unwrap();
```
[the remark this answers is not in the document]
schneems: I did this intentionally. I want both fixtures on disk and the ability to define them at runtime like this. I'm not sure I can entirely explain why I want it, but it's a thing I started doing on the legacy buildpack, and you can see I did in my Ruby Re-write as well:
-  https://github.com/heroku/buildpacks-ruby/blob/d526a49f81becaf571329a1adf5fff0668a6b99a/spec/cnb/buildpack_spec.rb#L12-L30
- https://github.com/heroku/heroku-buildpack-ruby/blob/a1dc09a93fc81a73482d6db0ecb02298001d5af4/spec/hatchet/rubies_spec.rb#L19-L36
- https://github.com/heroku/heroku-buildpack-ruby/blob/a1dc09a93fc81a73482d6db0ecb02298001d5af4/spec/hatchet/ruby_spec.rb#L129-L147

Some benefits of inlining the fixture:

- It's visible along with the code. No need to tab between the two. That's probably my biggest personal benefit.
- Total test isolation, no worries updating a fixture for one test will break another.

buildpacks/ruby/src/test_helper.rs
```diff
@@ -0,0 +1,14 @@
+use std::path::PathBuf;
+
+pub fn touch_file(path: &PathBuf, f: impl FnOnce(&PathBuf)) {
+    if let Some(parent) = path.parent() {
+        println!("path {path:?}");
+        println!("parent {parent:?}");
```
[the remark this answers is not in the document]
edmorley: Resolved in 7859f4f9ac4a5cf42726793b93c92bc0ca17f2c3

edmorley (approved):
Thank you for all the changes! I have no blockers to this landing now :-)

(I haven't looked too closely at some of the business logic [more the things around the edges], since I presume you/tests/others reviewing have covered those more [and things can be iterated on later on if needed anyway].)

Malax:
I'll defer my review since @edmorley and @joshwlewis already reviewed and approved - I trust these guys :) If you want a separate review from me just let me know. :)

commons/src/layer/mod.rs
```diff
@@ -0,0 +1,10 @@
+mod configure_env_layer;
+mod default_env_layer;
+
+#[allow(clippy::module_name_repetitions)]
+#[allow(clippy::useless_attribute)]
```
edmorley, line 5: The `clippy::useless_attribute` lint was warning about a valid issue, so I don't think it should be disabled.

The reason the `clippy::module_name_repetitions` lint was showing as "useless" was because it was missing the exclamation mark, so was being applied as an attribute an the individual `use` item, not as an attribute on the whole module.

See:
https://rust-lang.github.io/rust-clippy/master/index.html

.gitignore
```diff
@@ -6,3 +6,4 @@ repos/*
 *.gemspec
```
edmorley, line 6: I've just noticed there are a few Ruby entries in `.gitignore` that can be removed now :-)

ie: Pretty much all entries aside from `.DS_Store`, `target/` and `tmp/`

commons/src/layer/mod.rs
schneems: It's needed to get over this https://github.com/rust-lang/rust-clippy/issues/9401.

.gitignore
schneems: ❤️

commons/src/layer/mod.rs
schneems: Ahh I see what you're saying. This can apply to the whole mod file instead:

```
#![allow(clippy::module_name_repetitions)]
mod app_cache;
mod app_cache_collection;
mod clean;
mod config;
mod error;
mod in_app_dir_cache_layer;

pub use crate::cache::app_cache::AppCache;
pub use crate::cache::app_cache::CacheState;
pub use crate::cache::app_cache::{build, PathState};
pub use crate::cache::app_cache_collection::AppCacheCollection;
pub use crate::cache::clean::FilesWithSize;
pub use crate::cache::config::CacheConfig;
pub use crate::cache::config::{mib, KeepPath};
pub use crate::cache::error::CacheError;
```

I'm also happy to revisit the naming scheme, but it's quite confusing for me to not have the main API also be in a file with that name. If I want to find `CacheError` It's in `cache/error.rs`. To make clippy happy I would have to not use `Error` at the beginning or the end.

edmorley: > Ahh I see what you're saying. This can apply to the whole mod file instead:

Yeah exactly -- elsewhere in other files (and repos) where we've disabled the `clippy::module_name_repetitions` lint, we've done so globally, not per-statement, so it seems fine to disable globally here too.

PR #120, opened 2023-02-07
Refactor EnvCommand and introduce WhichProblem

schneems:
This is a PR intended to get implementation feedback. I want to replace `EnvCommand` with something more modular. To do that I introduced `commons::cmd` which is essentially a bag of related functions that compose well with each other.

Before re-writing the entire codebase I chose to rewrite only one command invocation inside of `bundle_install_layer.rs` to get a feel for the impacts.

Building on top of that work, I want to see what it looks like to annotate failed commands by default with `which_problem`.

I'm looking for feedback on:

- naming
- patterns
- differing philosophies/assumptions
- any other nitpicks or problems
- who would win in a legal fight: she hulk or birdman attorney at law?


buildpacks/ruby/src/layers/bundle_install_layer.rs
```diff
@@ -353,9 +355,14 @@ fn bundle_install(env: &Env) -> Result<(), CommandError> {
         ],
     );
 
-    user::log_info(format!("\nRunning command:\n$ {command}"));
+    user::log_info(format!("\nRunning command:\n$ {command_string}"));
```
schneems, line 358: I was thinking about making this chainable somehow.  Maybe use an extension trait on Command to implement some self yielding methods where we can add logging. For example `cmd_yield` where it yields itself but then returns itself.

buildpacks/ruby/src/layers/bundle_install_layer.rs
```diff
@@ -353,9 +355,14 @@ fn bundle_install(env: &Env) -> Result<(), CommandError> {
...
 
-    command.stream()?;
+    command
+        .output_and_write_streams(std::io::stdout(), std::io::stderr())
+        .map_err(|error| cmd::annotate_which_problem(error, bundle.into(), env.get("PATH")))
+        .map_err(cmd::os_command_error)
+        .and_then(|output| cmd::check_non_zero(output, cmd::OutputState::AlreadyStreamed))
```
schneems, line 364: We could provide two functions here for the variants and then you can call them without arguments.

buildpacks/ruby/src/layers/bundle_install_layer.rs
```diff
@@ -353,9 +355,14 @@ fn bundle_install(env: &Env) -> Result<(), CommandError> {
...
+        .map_err(|error| CmdError::new(command_string, error))?;
```
schneems, line 365: One of the hard things to solve about the ergonomics here is that I want to be able to chain/compose behavior. In order to get the error command in the error message, there's not a good way to do that since one of the core design goals is to interop natively with Command which returns either an Output or Child when run.

An additional API might be providing a function that takes a Command and then a config object like:

```
cmd::run(Command::new("bundle"), cmd::Config::default())?
```

Or as an extension:

```
Command::new("bundle").run_with(cmd::Config::default))?
```

Have it Return `Result<Option, CmdError>`.

We could still expose a composable API but with something like that, we could have a batteries-included experience. Make it what we want, and if it doesn't meet someone else's needs, they can still use the building blocks.

If we wanted to take functions as config, then the config could include:

- Function to run (Receives a `Command`, returns an `Output`)
- Function to log (Implement a default, but allow an over-ride)
- Function to convert Option into `Result<Option, CmdError>` (default to checking status code)
- Function to apply to std::io::Error (default to WhichProblem)


We have to take into consideration that Command may or may-not contain all the information that is needed in one of these. In addition to program name and arguments, it also takes environment variables and can potentially take a current directory https://doc.rust-lang.org/std/process/struct.Command.html#method.get_envs. I would need to experiment with the API to have more of an opinion.

I like the idea of giving out building blocks, but also having an opinionated batteries-included API.

buildpacks/ruby/src/layers/bundle_install_layer.rs
schneems: Would need two functions:

- `check_non_zero_already_streamed`
- `check_non_zero_include_output_in_error`

It's a bit of a mouthful. Not sure If I like it or not. Being explicit here seems nice. I don't think this is an optimization we need to make right now.

schneems:
Here's another experiment https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=decd260cddc65598c2a1f8ad5d6ea422

schneems:
Closing to preserve commits as they're referenced from other places, moving work over to another branch

PR #129, opened 2023-03-09
Use formatdoc! instead of format! in errors

schneems:
- Use formatdoc! instead of format! in errors

We don't want to emit the extra whitespace here. Use `formatdoc!` macro instead of `format!` (by mistake).

This would be a hard one to catch. I wonder if there's a way to detect/warn on it somehow. Maybe with clippy and a custom rule?

- Add version to changelog and document. It's unintuitive why the first release was 2.0.0 so I added a note.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
-## Unreleased
+## [Unreleased]
+
+## [2.0.0] 2023/31/01
 
 - Initial version of Ruby buildpack in Rust (https://github.com/heroku/buildpacks-ruby/pull/93)
+- Version 2.0.0 for the first release is not a typo. There was an issue in pack where a builder with the same name and version number would reuse artifacts left on image from [prior runs which caused issues](https://github.com/buildpacks/pack/issues/1322#issuecomment-1038241038). There were prior releases of `heroku/ruby` CNB from different sources that triggered this problem. To ensure no one would encounter that issue we developed and released using a version we know has not been used before. Version 2.0 was the first major version without a prior release of `heroku/ruby` CNB from any source.
--- a/README.md
+++ b/README.md
@@ -18,6 +18,37 @@ This buildpack is a Cloud Native Buildpack (CNB). CNB is a specification for bui
 
 You can find out more about the CNB project at https://buildpacks.io/.
 
+## Use `heroku/ruby` CNB locally with `pack`
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,4 +2,7 @@
 
 ## Unreleased
 
+## 2.0.0 - First release
+
 - Initial version of Ruby buildpack in Rust (https://github.com/heroku/buildpacks-ruby/pull/93)
+- Version 2.0.0 for the first release is not a typo. There's an issue in pack where a builder with the same name and version number would reuse artifacts left on image from prior runs. Previously a `heroku/ruby` CNB was released with older versions. That buildpack was based on Ruby and needed to bootstrap a copy of Ruby meaning that future runs of any `heroku/ruby` CNB with the same versions would also contain that bootstrapped Ruby binary. This causes confusion and difficult to debug problems. The way to ensure no one would encounter that issue was to release using a version we know has not been used before. Version 2.0 was the first major version without a prior release of any `heroku/ruby` named CNB.
```
edmorley, line 8: The pack issue was fixed year ago fwiw:
https://github.com/buildpacks/pack/issues/1322#issuecomment-1038241038

edmorley:
> This would be a hard one to catch. I wonder if there's a way to detect/warn on it somehow. Maybe with clippy and a custom rule?

This is why I assert against multi-line log output in my integration tests for the Python CNB :-)

schneems:
> This is why I assert against multi-line log output in my integration tests for the Python CNB :-)

I hear you. I prefer to not keep a 1:1 mapping of integration tests to every possible error case my app can generate. (I didn't mess up all of them, just some. I like where you're mind's at though!

I think I'll chalk this up to an "oops and we should watch out for it in the future".

If it ever shows up again in prod, we could maybe try a unit test against rendering all error cases and check that there's no newlines followed by a bunch of whitespace. The only difficult/annoying part would be generating the input error structs.

CHANGELOG.md
schneems: Updated. Also switched to the CHANGELOG header style in jvm and nodejs buildpacks.

schneems:
>  The only difficult/annoying part would be generating the input error structs

One more thought here: AFAIK there's no way to make sure that you've exhaustively generated every variant of an enum. So someone could add a new, untested variant, forget to add tests, and we wouldn't get an error/warning.

edmorley (approved):

PR #130, opened 2023-03-09
Add experimental support for skipping Ruby install

schneems:
Allows user or another buildpack to set `HEROKU_USE_SYSTEM_RUBY`. This will:

- Skip our logic for installing ruby
- Trigger a `which ruby` command to run, if it returns non-zero exit the build will not proceed
- Trigger a `ruby -v` command to run, if it returns non-zero exit the build will not proceed
- Includes a warning message stating that turning this on effectively disables any guarantees of support

The main driver is allowing for experimentation of replacing a Ruby binary.


jkutner:
Can you set:

```
[[stacks]]
id = "*"
```

buildpacks/ruby/src/ruby_version.rs
```diff
@@ -0,0 +1,208 @@
...
+/// ## Errors
+///
+/// - `HEROKU_USE_SYSTEM_RUBY` is set, but `ruby -v` errors
+/// - Ruby engine version is specified, without a Ruby version
+pub(crate) fn from_lockfile(lockfile: &str, env: &Env) -> Result<Ruby, RubyVersionError> {
+    if let Some(value) = env.get("HEROKU_USE_SYSTEM_RUBY") {
+        user::log_warning(
+            "Using System Ruby",
+            formatdoc! {"
+                Found environment variable {HEROKU_USE_SYSTEM_RUBY}={value:?}. When enabled, you are
```
schneems, line 38: As we're brainstorming an interface. One thought/question here is debug-ability. 

I'm calling `ruby -v` and erroring if I didn't find a Ruby version on disk. To help debug if that fails I can use https://github.com/schneems/which_problem (perfect use case for that logic TBH). However there's no visibility into where the user was expecting the Ruby version to come from. I think they would want to know:

- Did a user set `HEROKU_USE_SYSTEM_RUBY` or was it a buildpack (or host OS)? 
  - We can detect if it came from platform env or in current env
- If it came from buildpack/host-os how can we forward that location to the end user?
  - One idea would be to have another env var like `HEROKU_USE_SYSTEM_RUBY_REASON` and make it mandatory when not set via platform/user env. Presumably if it's set via another buildpack it would be something like `HEROKU_USE_SYSTEM_RUBY_REASON="core-os-ruby buildpack is installing Ruby"`
  - We could just use the value of `HEROKU_USE_SYSTEM_RUBY`, but I bet most buildpacks that choose to set this would just set it to 1 out of laziness. Though maybe not? We could get wild and enforce that it's semi-structured data like `buildpack: <name-of-buildpack>` or `os: <name-of-os-expected-to-have-ruby>`. I'm just trying to think of how we could give users more of a hint of where to look next when things go wrong other then "stuff failed"

It's useful to think about this now, as it would be a breaking change if we mandate something explicit/different in the future.

schneems:
@jkutner done. Also for completeness sake adding this link regarding stacks going away https://github.com/buildpacks/rfcs/issues/219.

I refactored a bunch of logic too. I'm quite happy with the end results. If we end up not keeping this experiment I'll likely want to keep the refactor minus the `HEROKU_USE_SYSTEM_RUBY` logic.

PR #155, opened 2023-06-28
Updated build output

schneems:
This is a proposed output and a spike on some APIs. For the sake of not having to re-edit the PR description, I'll post updates on the output below.

## Links

This PR is the result of quite a long process of revising the build output started around April 18th. Unfortunately, the documentation for those pieces is internal, but towards the end, I hope to release some formalized/finalized version. 

Primary documents:

- [Reverse timeline of build output](https://salesforce.quip.com/voJfA6oIVhrk) a chronological ordering of events regarding the build output. I started this doc on April 24 which as of today is 36 working days accounting for hackweek (5 days), my 10 year anniversary (5 days), and weekends (18 days). I've made 19 entries or a little under an entry every two days that I've worked on this project. Of all the documents I've created, this has been the most useful.
- [Build output spec (in progress)](https://docs.google.com/document/d/1_-DpPTGXaJBU1k7AJrjrNTp3iGL08yFnhrKSxVZrka0/edit) and related [output spike](https://docs.google.com/document/d/1lCceU6VRikwGRpCMMJtI1sU0GQXCMX64_Yq-OphzGzc/edit) which were used to generate the initial output we were settled on implementing.
- [CNB Build output META - COIL](https://salesforce.quip.com/aAkDAceUbdBb#EZaAAArbL63) a table-of-contents of sorts for the larger project.

Miscellaneous supporting documents:

- [Problems Values Stakeholders Possibilities Evaluation - Logging](https://salesforce.quip.com/GvlkA9vRJtka).
- [literature review](https://salesforce.quip.com/9HoIAfdHuGGe).
- [Why don't animations work in build outputs?](https://salesforce.quip.com/3fCRASFPiUCu) animations can be simulated by using codes to delete and re-write characters over again. This works with the CLI but not via `git push heroku`, and the doc details an exploration into the details of why.
- [CNB build output implementer problems](https://salesforce.quip.com/8BllApPUnlhI).
- [Ruby CNB build output for getting started guide 2023-04-18](https://salesforce.quip.com/eGekAZRAnMtf).
- [Historical: 2020/10/09 RFC: Heroku CNB Logging Style](https://salesforce.quip.com/rImvAvVyuulB).
- [Hackweek team docs](https://salesforce.quip.com/lVW7A6iGjQL9).
- [Node.js CNB Output Thoughts - Josh Lewis](https://salesforce.quip.com/0YZ8AGNTaLTC).
- [build output notes](https://salesforce.quip.com/RDW0Az7HvubZ) random assorted thoughts.
- [CNB build output actions](https://salesforce.quip.com/MTaoA9FXgWW5) a list of actionable activities for anyone who wants to participate in the exploration.
- [CLI Output Examples for Other SF Products](https://salesforce.quip.com/2gNQAWZPaG1x).
- [2023-05-15 Output spike (Ruby build)](https://salesforce.quip.com/or3bAJNEHiJc) replaced by [output spike](https://docs.google.com/document/d/1lCceU6VRikwGRpCMMJtI1sU0GQXCMX64_Yq-OphzGzc/edit#heading=h.cbtsbxcqv5rt).
- [CNB build output - User Problem collection](https://salesforce.quip.com/LxEBAWz3D5wN).
- [Ruby Classic build output for getting started guide 2023-04-18](https://salesforce.quip.com/XCEqA3uYhqZj).

Work item: GUS-W-13115766


schneems:
## 2023-06-28

Screenshot:

Screenshot 2023-06-28 at 4 23 09 PM

Raw output logs with no color:

```
$ docker rmi my-image --force  &&
pack build my-image --buildpack target/buildpack/debug/heroku_ruby --path buildpacks/ruby/tests/fixtures/default_ruby
Error: No such image: my-image
22: Pulling from heroku/builder
Digest: sha256:6f1ce423a17b35fec1a7863b3d100edec221470e2900680f6ead4c991f49bffd
Status: Image is up to date for heroku/builder:22
22-cnb: Pulling from heroku/heroku
Digest: sha256:4344cb4c208070615e278ebcb89fc5e47b3df2ea65492e92a8a29028e34e71d5
Status: Image is up to date for heroku/heroku:22-cnb
===> ANALYZING
Image with name "my-image" not found
===> DETECTING
heroku/ruby 2.0.0
===> RESTORING
===> BUILDING

# Heroku Ruby Buildpack

- Ruby version `3.1.3` from `default`
  - Installing [----------------] (15.718s)
- Bundler version `2.4.5` from `default`
- Bundle install
  - Running `BUNDLE_BIN="/layers/heroku_ruby/gems/bin" BUNDLE_CLEAN="1" BUNDLE_DEPLOYMENT="1" BUNDLE_GEMFILE="/workspace/Gemfile" BUNDLE_PATH="/layers/heroku_ruby/gems" BUNDLE_WITHOUT="development:test" bundle install`

      Fetching gem metadata from https://rubygems.org/..
      Fetching rake 13.0.6
      Installing rake 13.0.6
      Using bundler 2.4.5
      Fetching rack 2.2.3
      Fetching webrick 1.7.0
      Installing webrick 1.7.0
      Installing rack 2.2.3
      Bundle complete! 3 Gemfile dependencies, 4 gems now installed.
      Gems in the groups 'development' and 'test' were not installed.
      Bundled gems are installed into `/layers/heroku_ruby/gems`

  - Done (9.433s)
- Setting default processes(es)
  - Detecting gems
  - Running bundle list  [-] (0.143s)
  - Detected rack app (`rack` gem and `config.ru` at root of application)
- Rake assets install
  - Cannot run rake tasks (no Rakefile)
  ! Help: Add `Rakefile` to your project to enable
- Done (finished in 26.654s)
===> EXPORTING
Adding layer 'heroku/ruby:bundler'
Adding layer 'heroku/ruby:env_defaults'
Adding layer 'heroku/ruby:gems'
Adding layer 'heroku/ruby:ruby'
Adding layer 'buildpacksio/lifecycle:launch.sbom'
Adding 1/1 app layer(s)
Adding layer 'buildpacksio/lifecycle:launcher'
Adding layer 'buildpacksio/lifecycle:config'
Adding layer 'buildpacksio/lifecycle:process-types'
Adding label 'io.buildpacks.lifecycle.metadata'
Adding label 'io.buildpacks.build.metadata'
Adding label 'io.buildpacks.project.metadata'
Setting default process type 'web'
Saving my-image...
*** Images (670b4a5562dd):
      my-image
Reusing cache layer 'heroku/ruby:bundler'
Reusing cache layer 'heroku/ruby:gems'
Reusing cache layer 'heroku/ruby:ruby'
Successfully built image my-image
```

schneems:
> i guess i was looking for a design that mapped more to specific UX language/terminology.

One of the hardest things in computer science. I'm open to suggestions. I wanted to start with a broad suite of tools and get more specific over time. The `say_with_details` is one example there where I found I kept having to `build_output::fmt::details` again and again, and it made sense to have a helper API. Some of the argument names are slightly suggestive like `build_output::section()` takes a `topic` argument to suggest a noun rather than a verb.

I'm happy to brainstorm/sketch either naming or ideas in person or async. I had considered something like encoding a set of known/approved verbs i.e. installing/creating/downloading etc. to standardize on some of the languages and also give us a very consistent visual output, however, at this stage didn't want to be overly restrictive

commons/src/build_output.rs
```diff
@@ -0,0 +1,647 @@
...
+
+    /// Used to decorate a buildpack
+    #[must_use]
+    pub(crate) fn header(contents: impl AsRef<str>) -> String {
+        let contents = contents.as_ref();
+        colorize(HEROKU_COLOR, format!("\n# {contents}"))
+    }
+
+    /// Used to standardize error/warning/important information
+    pub(crate) fn lookatme(
```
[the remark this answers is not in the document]
schneems: It's an internal helper api. When I wrote it the voice in my head was using one breath like someone would with a tantruming toddler. Happy to change it.

commons/src/build_output.rs
```diff
@@ -0,0 +1,647 @@
...
+    println!("- {topic}");
+
+    section::Section { topic }
+}
+
+/// Top level buildpack header
+///
+/// Should only use one per buildpack
+#[must_use]
+pub fn header(buildpack: impl AsRef<str>) -> BuildpackDuration {
```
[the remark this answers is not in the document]
schneems: I did want something more specific and like that. 

I've been thinking about how this will apply to buildpacks and, specifically, multi-buildpacks. I.e. when the Node buildpack executes, do we want it to advertise each-and-every engine? One UX issue with he current output from Ruby's perspective is that there's no indication why the node engine is running in the body of the build output. One thought I had was if we could communicate with the dependency somehow and get some output like:

```
# Heroku Node Engine (for `heroku/ruby`)
...
```

buildpacks/ruby/src/layers/bundle_download_layer.rs
```diff
@@ -113,14 +106,14 @@ impl Layer for BundleDownloadLayer {
             version: self.version.clone(),
         };
         match cache_state(old.clone(), now) {
-            State::NothingChanged(version) => {
-                user::log_info(format!("Using bundler {version} from cache"));
+            State::NothingChanged(_version) => {
+                self.build_output.say("Using cached version");
```
joshwlewis, line 110: I'm not a huge fan of `say` here. Why not `log` or `info`?

commons/src/build_output.rs
```diff
@@ -0,0 +1,709 @@
...
+        }
+
+        /// Finish without announcing anything
+        #[allow(clippy::unused_self)]
+        pub fn done_silently(self) {}
+    }
+
+    /// Handles outputing inline progress based on timing
+    ///
+    /// i.e.   `- Installing [------] (5.733s)`
```
joshwlewis, line 99: I've seen other tools use ellipsis for this kind of stuff. For example:

`Installing.`
`Installing..`
`Installing...`

Even our own cli does a variation of this. What do you think of that?

commons/src/build_output.rs
```diff
@@ -0,0 +1,709 @@
...
+    }
+
+    impl Section {
+        /// Emit contents to the buid output with indentation
+        pub fn say(&self, contents: impl AsRef<str>) {
+            let contents = contents.as_ref();
+            println!("{SECTION_PREFIX}{contents}");
+        }
+
+        pub fn say_with_details(&self, contents: impl AsRef<str>, details: impl AsRef<str>) {
```
joshwlewis, line 271: Given details are just messages wrapped in parentheses which are super easy for authors to use, I'm not sure this has a lot of value. If I want to put details mid-sentence, then I'd have to drop down to using fmt::Display or `(` literals myself. Honestly, I'd probably just use literals.

joshwlewis (approved):
I left some comments. These are all non-blocking. Mine are all from the perspective of what I would want if this was a buildpack I work on regularly, or these functions were part of a shared toolkit across our buildpacks. For the Ruby buildpack, I think this looks good and is a substantial improvement.

commons/src/build_output.rs
schneems: This method is an extraction from a pattern I was doing over and over. You can manually format it with details, but it's a bunch of boilerplate:

```rust
let details = build_output::fmt::details("bundler version changed");
self.build_output
    .say(format!("Clearing cache {details}");
```

Versus this method needs no `format!` invocation and takes 2 lines instead of 3 (I also think it reads better):

```rust
self.build_output
    .say_with_details("Clearing cache", "bundler version changed");
```

commons/src/build_output.rs
```diff
@@ -0,0 +1,716 @@
...
+        Silent,
+        InlineProgress,
+    }
+}
+
+pub mod fmt {
+    use indoc::formatdoc;
+    use std::fmt::Display;
+
+    pub(crate) const RED: &str = "\x1B[31m";
```
joshwlewis, line 474: You may already know this, but this will choose red, without overriding the font weight. It's possible that this would render bold red if something else set the weight to bold. If you want to explicitly make this not-bold, not-italic, not-underlined with a red foreground, you could use `\x1B[0;31m`.

commons/src/build_output.rs
schneems: Pack uses `[=====>     ]` (but then clears them after they're done) here's an example:

image

Here's an example using dimmed dots:

image


And the raw logs (without colors):

```
$ $(set -pipefail; cd buildpacks/ruby && cargo libcnb package; cd ../..) &&
docker rmi my-image --force  &&
pack build my-image --buildpack target/buildpack/debug/heroku_ruby --path buildpacks/ruby/tests/fixtures/default_ruby
INFO - Reading buildpack metadata...
INFO - Found buildpack heroku/ruby with version 2.0.0.
INFO - Determining automatic cross-compile settings...
INFO - Building binaries (x86_64-unknown-linux-musl)...
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on build directory
   Compiling commons v0.1.0 (/Users/rschneeman/Documents/projects/work/buildpacks/buildpacks-ruby/commons)
   Compiling heroku-ruby-buildpack v0.0.0 (/Users/rschneeman/Documents/projects/work/buildpacks/buildpacks-ruby/buildpacks/ruby)
    Finished dev [unoptimized] target(s) in 17.90s
INFO - Writing buildpack directory...
INFO - Successfully wrote buildpack directory: ../../target/buildpack/debug/heroku_ruby (12.65 MiB)
INFO - Packaging successfully finished!
INFO - Hint: To test your buildpack locally with pack, run: pack build my-image --buildpack ../../target/buildpack/debug/heroku_ruby --path /path/to/application
Untagged: my-image:latest
Deleted: sha256:2e03f8025ffa454017f63cbb64b429efbaa479dcb35f2ea88d98fc8ebd7834b6
Deleted: sha256:d16e0fea83aa8c5c893878d8b7d1509b2449ce005e97bb359e44a8eceddd3ea9
Deleted: sha256:2235028ba88b6bdccbbdf56f3403c3d13731b253a227ec4e72674d34292598f3
Deleted: sha256:1578fe5946dbccf5ce0ed470766435a9136c5f5d48835eca47833536135800a3
Deleted: sha256:9e6c7628a8879594e2cd77dd73be3094f06f2ff11ee22942618e7fd916b8bf3b
Deleted: sha256:21545dba3c274ca7b7cb19ae6e3f4d5a96b9fb751295e4f4791d0165530f2dfe
Deleted: sha256:e1f001d2d22860ead8c04b64fd17428bc21d36260e248d4e101447986436482f
Deleted: sha256:043aa6e8e6e2cfecd2d91e16e00b80ad6159f9030d495eeceb694ef68735a05f
Deleted: sha256:11655cfc99cc816b9235e4db5770dc9e6f73254003f1d79502b18fa3cf425a08
Deleted: sha256:2b9692c22ea385e16af97ff3853a4c8930bc5bcd16db78532541120b219de69e
22: Pulling from heroku/builder
Digest: sha256:b3b0d2bb98243064670a4537ebc60bae0ebc669012bb6718285f1e3cf8768c44
Status: Image is up to date for heroku/builder:22
22-cnb: Pulling from heroku/heroku
Digest: sha256:726e37269cbdcd8d03499fbc6c9bf3d6fa7cf0b910ab2db0f3418ddc85407a60
Status: Image is up to date for heroku/heroku:22-cnb
===> ANALYZING
Image with name "my-image" not found
===> DETECTING
heroku/ruby 2.0.0
===> RESTORING
===> BUILDING

# Heroku Ruby Buildpack

- Ruby version `3.1.3` from `default`
  - Installing.......(4.671s)
- Bundler version `2.4.5` from `default`
  - Running `gem install bundler --version 2.4.5`....(1.328s)
- Bundle install
  - Running `BUNDLE_BIN="/layers/heroku_ruby/gems/bin" BUNDLE_CLEAN="1" BUNDLE_DEPLOYMENT="1" BUNDLE_GEMFILE="/workspace/Gemfile" BUNDLE_PATH="/layers/heroku_ruby/gems" BUNDLE_WITHOUT="development:test" bundle install`

      Fetching gem metadata from https://rubygems.org/..
      Fetching rake 13.0.6
      Installing rake 13.0.6
      Using bundler 2.4.5
      Fetching rack 2.2.3
      Fetching webrick 1.7.0
      Installing webrick 1.7.0
      Installing rack 2.2.3
      Bundle complete! 3 Gemfile dependencies, 4 gems now installed.
      Gems in the groups 'development' and 'test' were not installed.
      Bundled gems are installed into `/layers/heroku_ruby/gems`

  - Done (4.622s)
- Setting default processes(es)
  - Detecting gems
  - Running `bundle list`...(0.208s)
  - Detected rack app (`rack` gem and `config.ru` at root of application)
- Rake assets install
  - Cannot run rake tasks (no `Rakefile`)
  ! Help: Add `Rakefile` to your project to enable
- Done (finished in 10.927s)
```

buildpacks/ruby/src/layers/bundle_download_layer.rs
schneems: One intention is trying to distinguish between "logging" which universally everyone (not on this team) assumes means something that the user will never see versus "emit to build output" which is clear that it's intended for the end user. To that end, I'm trying to convey "you are communicating with the user now".

- `log` I want to avoid implying this is a non-visible log
- `info` makes it log-like and makes it seem like there's also `debug` or `fatal` and that that level can be configured. I find that style of "log level" confusing and want to avoid introducing it.

Brainstorming a bit. We could call it:

- out
- output
- println (want to avoid so people don't accidentally use `println!` by mistake)
- emit
- report
- msg
- message
- note (though possibly confusing as they might assume it means `Note: {text}` will be output
- write
- line
- bulletpoint

Or maybe a synonym https://www.thesaurus.com

commons/src/build_output.rs
```diff
@@ -0,0 +1,716 @@
...
+            .join("\n")
+    }
+
+    /// Colorizes a body while preserving existing color/reset combinations and clearing before newlines
+    ///
+    /// Colors with newlines are a problem since the contents stream to git which prepends `remote:` before the `libcnb_test`
+    /// if we don't clear, then we will colorize output that isn't ours.
+    ///
+    /// Explicitly uncolored output is handled by a hacky process of treating two color clears as a special case
+    pub(crate) fn colorize(color: &str, body: impl AsRef<str>) -> String {
```
joshwlewis, line 665: All this search and replace is a bit worrying. There are many ways this could go wrong. I'd drop all these `replace` calls. If someone wants to ensure some text is uncolored, they can use `/x1b[39m`, which specifies default foreground color. Then, wrapping with red won't override it.

schneems: Your comments on escape codes Are really helpful. This is my main resource https://sweworld.net/linux/terminal/escape_code/ 

This is a great suggestion, and it clears my hack:

>  \x1b[1;39m

But it makes me realize I guess I don't know what \x1b[0m is actually doing. I thought it was resetting things, but I don't know what would be different between that and "Default"

commons/src/build_output.rs
joshwlewis: Pack uses `[=====>     ]` (but then clears them after they're done)

Yeah, I like this style, but `pack` has the advantage of knowing it's in a terminal that supports clears. Output edits just don't work well in dashboard, and we should probably avoid it.

Personally, I think `Running `bundle list`...(0.208s)` looks great, though I'd probably add a space after the last dot. (like `Running `bundle list`... (0.208s)`).

I think dim text has fairly limited support in terminals, but they should fall back to standard text without issue.

commons/src/build_output.rs
schneems: Updated to remove the replace hack for NOCOLOR (which turns out wasn't even used). I'm still curious about the difference though (if you know)

Regarding `replace` usage, I couldn't find any off-the-shelf colorization libraries that handled the color-within-a-color. I'm game to update this to another approach. One idea I had was using `nom` to parse the incoming text and adjust it somehow. Or maybe I could try to grab some time with burnt sushi (who has significant experience with colorization). But I might put that off until we work to formalize the implementation. 

FWIW there's not a ton of tests in this new code, but this colorize function is tested. This method and the associated constants are all private and cannot be accessed except from within the module.

joshwlewis: You are right - `\x1b[0m` is a full reset. It clears all active styles - foreground color, background color, bold, dim, underline, blink (and others I'm probably forgetting). 

`\x1b[39m` sets the foreground color to the default color. I don't think `\x1b[39m` will fix all problems though. For example:

```
\x1b[31m]This is red\x1b[39mThis is default\x1b[0mThis is also default, even though we wanted to wrap it in red.\x1b[0m
```

With regard to `replace` usage, I was hoping to drop it completely rather than find something new. Maybe `colorize` shouldn't support color-in-color text - after all, this isn't really a thing ANSI escape codes support. ANSI escape codes don't have a tree-like structure by nature.

schneems: > Maybe colorize shouldn't support color-in-color text

If we can't have style-in-a-style then it would have the effect of forcing "warning", "error", and "important" to all be all one color. I think it's nice to have this ability, take the `! Help`  output for example, it's colorized to draw your attention, but then there are values inside of it. 

Right now there are no control codes that go into `colorize` that we don't own so I have a high degree of confidence in its implementation. One reason that's true is that currently, other build tools (like bundle install) are not being detected as a TTY. But we've not talked about that before:

From CNB execution https://github.com/heroku/buildpacks-ruby/actions/runs/5523980642/jobs/10075643688?pr=155

image

Running `bundle install` locally: 

image

Notice the green is present locally but not on CNB deploys.

I think this CNB behavior matches classic behavior, but it does ask the question: What output do we want? Should our subcommands have native colorization or be plain?

From an implementation standpoint, I think supporting colorized subcommands makes life harder (but not impossible). For this PR I don't want to change this behavior, but I do want to understand/explore our idealized outcome.

PR #157, opened 2023-06-29
Upgrade to spec 0.9 and libcnb 0.13.0

schneems:
Libcnb now targets spec version 0.9. The largest change is that bash processes no longer work for default processes. The workaround is to manually invoke `bash -c ` ([python example](https://github.com/heroku/buildpacks-python/blob/40d1abb9733f0a483c2a913ae88d41a41dd52277/src/salesforce_functions.rs#L57)).

---
Bumps [libcnb](https://github.com/heroku/libcnb.rs) from 0.11.5 to 0.13.0.
- [Release notes](https://github.com/heroku/libcnb.rs/releases)
- [Changelog](https://github.com/heroku/libcnb.rs/blob/main/CHANGELOG.md)
- [Commits](https://github.com/heroku/libcnb.rs/compare/v0.11.5...v0.13.0)

---
updated-dependencies:
- dependency-name: libcnb dependency-type: direct:production update-type: version-update:semver-minor ...


schneems:
I think we need to upgrade the procfile buildpack to 0.9 spec too. Otherwise, I have no clue how the two will play together. I think this is good to go though.

schneems:
https://github.com/heroku/procfile-cnb/issues/151

buildpacks/ruby/src/steps/get_default_process.rs
```diff
@@ -64,16 +64,33 @@ fn detect_web(gem_list: &GemList, app_path: &Path) -> WebProcess {
...
-        .args(["exec", "rackup", "--port", "$PORT", "--host", "0.0.0.0"])
+fn bashify(
+    program: impl Into<String>,
+    args: impl IntoIterator<Item = impl Into<String>>,
+) -> Process {
+    let args = args
+        .into_iter()
+        .map(std::convert::Into::into)
+        .collect::<Vec<String>>()
+        .join(" ");
```
Malax, line 75: This will produce unexpected results when an an arg contains spaces. Maybe https://docs.rs/shell-words/latest/shell_words/ is a good choice to add here. It wouldn't be a big problem when all `Process`es are created literally, but this helper "promises" to convert any program + args into a bash process. I'd argue not properly quoting is a bug.

Malax (changes requested):

buildpacks/ruby/src/steps/get_default_process.rs
schneems: There are some unintended side effects off just slapping that on there:

```
exec rackup --port '$PORT' --host 0.0.0.0
```

In this case I want it to be `"$PORT"` which gets evaluated instead of `'$PORT'` which is a string literal. We ran into this in the hatchet library as well where we ended up needing a way to specify "I actually want this to be evaluated". Maybe an enum:

```rust
enum ShellString {
 Escape(String), // shell escape whatever we pass in
 Quote(String), // ensure contents are wrapped in a doublequote but not shell escaped
 Raw(String), // Do nothing, pass whatever we're given, responsibility is on the person inputting the contents
}
```

Then the API could take an iterator of ShellString and force users to be explicit with how they want their args escaped (or not).

edmorley:
> Need to get cnb_shim inline with API spec 0.9 before procfile-cnb can be updated before this can be merged in [heroku/procfile-cnb#151](https://github.com/heroku/procfile-cnb/issues/151)

It's up to you, but I would say you could proceed with updating Ruby to the newer API even before cnb-shim is updated, since:
- some other buildpacks have already updated
- it only breaks people using cnb-shim with `profile.d`-creating legacy buildpacks (of which there aren't likely to be many at present)
- we're still in a pre-beta phase where edge cases like combining with shimmed buildpacks are mostly out of scope IMO

buildpacks/ruby/src/steps/get_default_process.rs
```diff
@@ -64,16 +64,86 @@ fn detect_web(gem_list: &GemList, app_path: &Path) -> WebProcess {
...
+impl Display for ShellString {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ShellString::Escape(string) => f.write_str(&shell_words::quote(string)), // single quote only if needed
+            ShellString::QuoteEnvVar(string) => write!(f, "\"{string}\""),
+        }
+    }
+}
+
+fn bashify(program: &ShellString, args: impl IntoIterator<Item = ShellString>) -> Process {
```
Malax, line 92: Food for thought: Do we even need this here? It's only being use twice in the whole code-base and the amount of a) code and b) cognitive overhead it introduces don't seem to be a worthwhile trade-off in my book - especially after we discovered we had to introduce another dependency for it.

I think the code would improve by removing `bashify` (and all supporting code) and just have literals for the two bash commands we have in the buildpack.

Malax (dismissed):
Request changes for the `Display` trait stuff. Feel free to ignore the quip about maybe not needing `bashify` at all - even tho I think that would be the better direction.

buildpacks/ruby/src/steps/get_default_process.rs
```diff
@@ -64,16 +64,86 @@ fn detect_web(gem_list: &GemList, app_path: &Path) -> WebProcess {
...
+    fn escape(arg: impl Into<String>) -> ShellString {
+        ShellString::Escape(arg.into())
+    }
+
+    fn quote_env_var(arg: impl Into<String>) -> ShellString {
+        ShellString::QuoteEnvVar(arg.into())
+    }
+}
+
+impl Display for ShellString {
```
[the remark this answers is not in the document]
schneems: > IMO, this is a misuse of the Display trait

I originally implemented to string and something told me I was wrong and should actually implement `Display`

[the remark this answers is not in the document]
schneems: Found it: https://doc.rust-lang.org/std/string/trait.ToString.html

> This trait is automatically implemented for any type which implements the [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait. As such, ToString shouldn’t be implemented directly: [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) should be implemented instead, and you get the ToString implementation for free.

buildpacks/ruby/src/steps/get_default_process.rs
schneems: I think any of the above are fine. Longer term I would ideally want libcnb/libherokubuildpack helper to make this a process a little nicer. 

I ended up reverting the "bashify" stuff and doing this manually. It's (perhaps) more error-prone as escaping must be remembered each time, but is less code and less toil. I honestly don't have a preference for either solution.

schneems:
Manuel is out. I'm going to ask @edmorley for a review. I just implemented this suggestion https://github.com/heroku/buildpacks-ruby/pull/157#discussion_r1254750159. This is a two-way door, and I can always set up a refactoring session with Manuel when he comes back if there are further style considerations to explore.

edmorley (approved):
Looks great! :-)

buildpacks/ruby/buildpack.toml
```diff
@@ -1,4 +1,4 @@
-api = "0.8"
+api = "0.9"
```
edmorley, line 1: Ah this needs a `CHANGELOG.md` entry, since it affects the tool versions that work with the buildpack (old Pack CLI won't work with the buildpack etc)

PR #172, opened 2023-07-19
Add Ruby app metrics support

schneems:
Ruby supports "language specific metrics (beta)" https://devcenter.heroku.com/articles/language-runtime-metrics-ruby through the barnes gem https://github.com/heroku/barnes and the `heroku/metrics` buildpack https://github.com/heroku/heroku-buildpack-metrics. This PR ports the logic of `heroku/metrics` to the Ruby CNB buildpack.

## Implementation

At build time, this buildpack will download a statsd client (named "agentmon") from S3 and place it on the path. There are currently no explicit interfaces for running background/daemon processes via the CNB spec, however there is an interface that guarantees it will execute arbitrary code (exec.d). This interface is intended to set dynamic environment variables at runtime. As a side effect, this interface is used to schedule a background daemon via the `start-stop-daemon` command [docs](https://www.man7.org/linux/man-pages/man8/start-stop-daemon.8.html).

The `start-stop-daemon` calls our program, a new rust script in `bin/agentmon_loop`. This rust script will boot agentmon in a loop if the necessary environment variables are set. 

GUS-W-13115697.
GUS-W-13815707.


buildpacks/metrics-agent/buildpack.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[[stacks]]
+id = "heroku-22"
+
+[[buildpack.licenses]]
+type = "BSD-3-Clause"
+
+[metadata]
+[metadata.release]
+[metadata.release.docker]
```
colincasey, line 22: the key has been renamed from `docker` to `image`

buildpacks/metrics-agent/tests/integration_test.rs
```diff
@@ -0,0 +1,33 @@
...
+use libcnb_test::{assert_contains, BuildConfig, BuildpackReference, ContainerConfig, TestRunner};
+use std::path::PathBuf;
+
+#[test]
+#[ignore = "integration test"]
+fn test_barnes_app() {
+    TestRunner::default().build(
+        BuildConfig::new("heroku/builder:22", "tests/fixtures/barnes_app").buildpacks(vec![
+            BuildpackReference::Crate,
+            BuildpackReference::Local(PathBuf::from("../ruby")),
```
colincasey, line 12: nice

schneems: Works like a charm.

buildpacks/metrics-agent/buildpack.toml
schneems: Thanks for the heads up. I also need to make a docker registry for this too.

buildpacks/metrics-agent/src/layers/install_agentmon.rs
```diff
@@ -0,0 +1,329 @@
...
+        // Sets executable bit
+        let mut mode = 0o455;
+        octal_executable_permission(&mut mode);
+        assert_eq!(0o755, mode);
+
+        // Does not affect already executable
+        let mut mode = 0o745;
+        octal_executable_permission(&mut mode);
+        assert_eq!(0o745, mode);
+    }
```
Malax, line 328: This essentially tests that bitwise OR works in Rust. I don't think we need that at all. There is only _one_ call-site for   `octal_executable_permission` which can easily replaced by inlining the `|= 0o700` making the function itself and the tests for it redundant.

buildpacks/metrics-agent/src/layers/install_agentmon.rs
```diff
@@ -0,0 +1,329 @@
...
+
+/// Ensures the provided octal number's executable
+/// bit is enabled.
+///
+/// i.e. chmod +x will ensure that the first digit
+/// of the file permission is 7 on unix so if you pass
+/// in 0o455 it would be mutated to 0o755
+fn octal_executable_permission(mode: &mut u32) {
+    *mode |= 0o700;
+}
```
Malax, line 281: This function just wraps a bitwise OR and has exactly one call-site. If we just inline the `|=` we can remove all that extra machinery, docs and tests. 

The docs for this function and `chmod_plus_x` make that even more evident since they both reference each others responsibility and blur into one. 

If we think we need to comment how bitwise operations work (I personally don't think we do) we can add a comment above the `|=` too.

buildpacks/metrics-agent/src/layers/install_agentmon.rs
```diff
@@ -0,0 +1,329 @@
...
+        .map_err(DownloadAgentmonError::CouldNotWriteDestinationFile)?;
+
+        execd_script
+    };
+
+    Ok(execd_script)
+}
+
+#[cached]
+fn get_latest_url() -> Result<Url, GetUrlError> {
```
Malax, line 214: We want to move to a more secure way of dealing with downloaded binaries for language runtimes and others. This implementation is not only excessively complex (cached macro, IO, URL parsing), it also doesn't verify he integrity of the download binary. 

The strategy implemented here also means that the behaviour at runtime could change or outright break, even when we don't release an updated buildpack. If a new agentmon version comes out, CI isn't run for it and customers immediately get the new version for their next build - including during moratoriums, etc.

I suggest that we add buildpack metadata for the URL and checksum of the most recent agentmon release which we can then use for downloading and verification. We can also add a simple automation that updates the `buildpack.toml` with the most recent version as a GHA that opens a PR for it.

This way we:
- Verify the binary, removing an attack vector
- Only explicitly update the agentmon version
- Enforce testing when a new agentmon version comes out
- Remove IO and therefore potential of errors at runtime
- Remove the need to memoize this function is a macro

buildpacks/metrics-agent/src/layers/install_agentmon.rs
```diff
@@ -0,0 +1,329 @@
...
+        <Self::Buildpack as libcnb::Buildpack>::Error,
+    > {
+        self.section
+            .say_with_details("Clearing cache", "invalid metadata");
+
+        Ok(libcnb::layer::MetadataMigration::RecreateLayer)
+    }
+}
+
+fn write_execd(agentmon_path: &Path, layer_path: &Path) -> Result<PathBuf, DownloadAgentmonError> {
```
Malax, line 156: `DownloadAgentmonError` seems wrong here. Either the name must change or the error should be specific to this function and wrapped at the call-site.

buildpacks/metrics-agent/src/layers/install_agentmon.rs
```diff
@@ -0,0 +1,329 @@
...
+    // We use the exec.d to boot a process. This script MUST exit though as otherwise
+    // The container would never boot. To handle this we intentionally leak a process
+    let execd_script = {
+        let execd_script = layer_path.join("agentmon_exec.d");
+        let log_file = layer_path.join("background.log");
+        fs_err::write(&log_file, "")
+            .map_err(DownloadAgentmonError::CouldNotCreateDestinationFile)?;
+
+        let log_file = log_file.display();
+        let agentmon_loop = agentmon_loop.display();
```
Malax, line 184: Nit: this made me pause for a moment - "how can the loop be displayed? And why it is re-assigned?" Renaming would go a long way here.

buildpacks/metrics-agent/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+        libcnb::Error::BuildpackError(err) => Cause::OurError(err),
+        err => Cause::FrameworkError(err),
+    }
+}
+
+pub(crate) fn on_error(err: libcnb::Error<MetricsAgentError>) {
+    match cause(err) {
+        Cause::OurError(error) => log_our_error(error),
+        Cause::FrameworkError(error) => ErrorInfo::header_body_details(
+            "heroku/buildpack-ruby internal buildpack error",
```
Malax, line 85: I don't understand why the additional `Cause` wrapper needed to be introduced here, the following seems to work just fine:

```suggestion
pub(crate) fn on_error(err: libcnb::Error<MetricsAgentError>) {
    match err {
        libcnb::Error::BuildpackError(error) => log_our_error(error),
        error => ErrorInfo::header_body_details(
            "heroku/buildpack-ruby internal buildpack error",
```

Maybe even better: using `libherokubuildpack::error::on_error`. If that is not good enough (the error message here is much more elaborate) we should change it there and let all other buildpacks reap the benefits too. :)

Malax (changes requested):
Cool to see progress on language metrics, I love that feature! I timeboxed the review to two hours and stopping now, I will do a second pass on this after the discussions I added in this review are resolved.

In general: This doesn't seem to be Ruby specific and could be of use for other languages as well - it might be worth extracting out at a later point in time. However, since this relies on the Ruby helpers in this repo it's already tied to it to some degree and will "calcify" more over time. It might therefore make sense to extract it now (or immediately after this PR is merged) to avoid it being long-term stuck in this repo.

buildpacks/metrics-agent/src/layers/install_agentmon.rs
schneems: I moved this to a function specifically so I could test it as my original implementation was incorrect. Working with bitwise operators is something I've done for school, but rarely in Rails work. The value of the test, and docs are that future programmers have that info at their fingers without looking it up.

buildpacks/metrics-agent/src/bin/agentmon_loop.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        match result {
+            Ok(status) => eprintln!("agentmon completed with status=${status}. Restarting"),
+            Err(error) => {
+                eprintln!("agentmon could not be run due to error: {error}");
+                eprintln!("Retrying");
+            }
+        };
+    } else {
+        eprintln!("PORT is not set, sleeping {sleep_for:?}");
```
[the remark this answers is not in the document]
schneems: It's technically possible. But I agree, very unlikely. Moving it makes things more consistent.

schneems:
I updated `agentmon_loop` I'm still looking at your other comments for main.rs and the layer probably tomorrow.

buildpacks/metrics-agent/src/layers/install_agentmon.rs
schneems: I've decided to inline it as ssuggested and add an extra assertion to the existing test:

```rust
        assert_eq!(after, after | 0o700);
```

buildpacks/metrics-agent/src/layers/install_agentmon.rs
schneems: This logic was a straight 1:1 port of the existing buildpack logic. The easiest thing to move forward would be to hardcode the path. Agentmon hasn't been updated in 6 years https://github.com/heroku/agentmon.

I would love some standardization on some of this stuff. Ideally being able to have some code like:

```rust
libcnb::MaifestDownload::new("agentmon.manifest")
  .with_tag(ManifestTag::KeyValue("stack", "heroku-22")
  .with_tag(ManifestTag::Flag("latest")
  .validate(ManifestValidate::Sha, ManifestTag::Key("sha"))
  .tar_gz()
  .download(layer_path)
  .map_err(...)
```

schneems: Updated to hardcode the value. Take a look, please.

schneems:
Updated to address the rest of the comments. I reworked a lot. Moved the bash script that schedules the daemon into a rust script. Moved to a hardcoded URL.

buildpacks/metrics-agent/src/bin/agentmon_loop.rs
```diff
@@ -0,0 +1,175 @@
...
+const SLEEP_FOR: Duration = Duration::from_secs(1);
+
+/// Agentmon Loop
+///
+/// Boots agentmon (a statsd server) in a loop
+///
+/// Example:
+///
+/// ```shell
+/// $ cargo run --bin agentmon_loop -- --path <path/to/agentmon/binary>
```
joshwlewis, line 24: Ugh, I wish we had a better way to keep a process running. Really `systemd` is almost the solution (it can start and keep a process running, and is on most linux distros), but it needs to be started under PID 1.

joshwlewis (commented):
This looks good to me; though I'd like to see the question about the generic buildpack resolved.

buildpacks/metrics-agent/src/layers/install_agentmon.rs
```diff
@@ -0,0 +1,287 @@
...
+        .unpack(destination.as_ref())
+        .map_err(InstallAgentmonError::CouldNotUnpack)
+}
+
+/// Sets file permissions on the given path to 7xx (similar to `chmod +x <path>`)
+///
+/// i.e. chmod +x will ensure that the first digit
+/// of the file permission is 7 on unix so if you pass
+/// in 0o455 it would be mutated to 0o755
+pub fn chmod_plus_x(path: &Path) -> Result<(), std::io::Error> {
```
edmorley, line 235: This doesn't need to be `pub` (or even `pub(crate)`) at the moment. Using `pub` means Clippy's deadcode detection won't work. (If you plan to move this into a common library in the future then make it public at that point. For now nothing can use it from here anyway.)

```suggestion
fn chmod_plus_x(path: &Path) -> Result<(), std::io::Error> {
```

buildpacks/metrics-agent/src/layers/install_agentmon.rs
```diff
@@ -0,0 +1,287 @@
...
+    let mut perms = fs_err::metadata(path)?.permissions();
+    let mut mode = perms.mode();
+    mode |= 0o700;
+    perms.set_mode(mode);
+
+    fs_err::set_permissions(path, perms)
+}
+
+pub(crate) fn download(
```
edmorley, line 244:
```suggestion
fn download(
```

buildpacks/metrics-agent/src/layers/install_agentmon.rs
```diff
@@ -0,0 +1,287 @@
...
+    download(url, agentmon_tgz.path())?;
+
+    untar(agentmon_tgz.path(), destination)?;
+
+    chmod_plus_x(&destination.join("agentmon")).map_err(InstallAgentmonError::PermissionError)?;
+
+    Ok(())
+}
+
+pub(crate) fn untar(
```
edmorley, line 219:
```suggestion
fn untar(
```

commons/src/build_output.rs
```diff
@@ -533,14 +533,15 @@ pub mod fmt {
         let noun = noun.as_ref();
         let header = header.as_ref();
         let body = help_url(body, url);
```
edmorley, line 535: These output changes seem unrelated? If so, I think they should be in their own PR :-)

edmorley (changes requested):

buildpacks/ruby/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- buildpacks/ruby: Introduce heroku build metrics support (https://github.com/heroku/buildpacks-ruby/pull/172)
```
edmorley, line 10: This changelog is specific to `buildpacks/ruby` - why is the `buildpacks/ruby` prefix needed here (there's nothing for it to differentiate against)?

buildpacks/ruby/Cargo.toml
```diff
@@ -23,6 +23,7 @@ tempfile = "3"
 thiserror = "1"
 ureq = "2"
 url = "2"
+clap = { version = "4.4.1", features = ["derive"] }
```
edmorley, line 26: The pattern we're been using (and can be seen in the lines above), is major versions only in `Cargo.toml`:

```suggestion
clap = { version = "4", features = ["derive"] }
```

edmorley (changes requested):
I took another look since it seemed like the previous review comments had been addressed, however, I see a number of the last review's comments are still open?
https://github.com/heroku/buildpacks-ruby/pull/172#pullrequestreview-1595381205

buildpacks/metrics-agent/src/bin/launch_daemon.rs
```diff
@@ -0,0 +1,77 @@
...
+
+fn main() {
+    let Args {
+        log,
+        loop_path,
+        agentmon,
+    } = Args::parse();
+
+    let mut command = Command::new("start-stop-daemon");
+    if std::env::var_os("AGENTMON_DEBUG").is_some() {
```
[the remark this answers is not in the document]
schneems: I'm updating it with instructions on how to disable the commit has more info on why the logic is the way it is i.e. it's intended that any value will enable it even if that value is `"false"`. I'll leave this unresolved to let you have a look.

buildpacks/metrics-agent/src/bin/agentmon_loop.rs
```diff
@@ -0,0 +1,175 @@
+// Enable Clippy lints that are disabled by default.
+// https://rust-lang.github.io/rust-clippy/stable/index.html
```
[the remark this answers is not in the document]
schneems: I want to hold myself to a higher standard by default. I'll leave it in unless it becomes an issue.

schneems:
I think there's a race condition in the tests previously it was failing with

```
thread 'test_barnes_app' panicked at 'assertion failed: `(left contains right)`
left (unescaped):
    PID TTY      STAT   TIME COMMAND
      1 ?        Rs     0:00 ps x
     14 ?        R      0:00 [start-stop-daem]


left (escaped): `"    PID TTY      STAT   TIME COMMAND\n      1 ?        Rs     0:00 ps x\n     14 ?        R      0:00 [start-stop-daem]\n"`
right: `"agentmon_loop --path"`', buildpacks/ruby/tests/integration_test.rs:138:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Now it's failing with

```
---- test_barnes_app stdout ----
    PID TTY      STAT   TIME COMMAND
      1 ?        Rs     0:00 ps x
     14 ?        R      0:00 start-stop-daemon --start --background --exec /layers/heroku_ruby/metrics_agent/agentmon_loop -- --path /layers/heroku_ruby/metrics_agent/bin/agentmon


thread 'test_barnes_app' panicked at 'assertion failed: `(left contains right)`
left (unescaped):
    PID TTY      STAT   TIME COMMAND
      1 ?        Rs     0:00 ps x
     14 ?        R      0:00 start-stop-daemon --start --background --exec /layers/heroku_ruby/metrics_agent/agentmon_loop -- --path /layers/heroku_ruby/metrics_agent/bin/agentmon


left (escaped): `"    PID TTY      STAT   TIME COMMAND\n      1 ?        Rs     0:00 ps x\n     14 ?        R      0:00 start-stop-daemon --start --background --exec /layers/heroku_ruby/metrics_agent/agentmon_loop -- --path /layers/heroku_ruby/metrics_agent/bin/agentmon\n"`
right: `"agentmon_loop --path"`', buildpacks/ruby/tests/integration_test.rs:[138](https://github.com/heroku/buildpacks-ruby/actions/runs/6153895381/job/16698442352?pr=172#step:8:139):21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I think it looks like this:

- start-stop-daemon is called, this can show up as one of the above outputs
- that will cause the `agentmon_loop --path <>`  command to be spawned, but maybe not right away

So depending on when we read the results of `ps` we can get a different result. I could possibly start the container and then loop/sleep a few times to try to get some determinisim. Maybe there's a better way to assert that:

- The script we want to boot does in fact boot at launch
- It does not immediately exit (which might indicate a problem in our code)

Another tool to use is to look at the specified output log file (if we set `AGENTMON_DEBUG=1`). It would probably be best   to print that out anyway in case an error comes up.

Cargo.toml
```diff
@@ -1,6 +1,10 @@
...
-]
+members = ["buildpacks/ruby", "commons"]
+
+[workspace.package]
+rust-version = "1.64"
+edition = "2021"
+license = "BSD-3-Clause"
+
+[workspace.dependencies]
+libherokubuildpack = "0.12"
```
edmorley, line 10: None of these entries are used (adding them to the root doesn't make them used, they need to be referenced from the individual crates too) - I presume they are leftover from when the feature was being added as it's own CNB?

If you want to switch the Ruby CNB + commons crates to sharing metadata via the root `Cargo.toml`, then that should be in its own PR :-)

buildpacks/ruby/src/bin/agentmon_loop.rs
```diff
@@ -0,0 +1,204 @@
+// Enable Clippy lints that are disabled by default.
+// https://rust-lang.github.io/rust-clippy/stable/index.html
+#![warn(clippy::pedantic)]
+
```
edmorley, line 4: (Informational)

I was curious how large these binaries were, so checked the release builds:

```
$ du -ah
2.6M    ./agentmon_loop
2.6M    ./launch_daemon
5.1M    .
```

I suspect much of that comes from Clap (and some of its default feature: https://docs.rs/clap/latest/clap/_features/index.html#default-features).

Though compared to the size of the Ruby runtime, Ruby dependencies and app source for the typical Ruby app, it's a rounding error, so nothing to worry about.

buildpacks/ruby/src/layers/metrics_agent_install.rs
```diff
@@ -0,0 +1,286 @@
...
+/// - Repo: <https://github.com/heroku/agentmon>
+/// - Releases: <https://github.com/heroku/agentmon/releases>
+///
+/// To get the latest s3 url:
+///
+/// ```shell
+/// $ curl https://agentmon-releases.s3.amazonaws.com/latest
+/// ```
+const DOWNLOAD_URL: &str =
+    "https://agentmon-releases.s3.amazonaws.com/agentmon-0.3.1-linux-amd64.tar.gz";
```
edmorley, line 28:
```suggestion
/// $ curl https://agentmon-releases.s3.us-east-1.amazonaws.com/latest
/// ```
const DOWNLOAD_URL: &str =
    "https://agentmon-releases.s3.us-east-1.amazonaws.com/agentmon-0.3.1-linux-amd64.tar.gz";
```

(For context, see https://github.com/heroku/agentmon/pull/18)

buildpacks/ruby/tests/integration_test.rs
```diff
@@ -112,6 +112,36 @@ fn test_ruby_app_with_yarn_app() {
         );
 }
 
+#[test]
+#[ignore = "integration test"]
+fn test_barnes_app() {
+    TestRunner::default().build(
+        BuildConfig::new("heroku/builder:22", "tests/fixtures/barnes_app")
+            .buildpacks(vec![BuildpackReference::Crate]),
```
edmorley, line 120: `BuildpackReference::Crate` is the default, so this can be omitted

schneems:
I fixed the tests (https://github.com/heroku/buildpacks-ruby/pull/172#issuecomment-1714857517) to be **more** deterministic, but it's still based on a timeout (which can fail if the machine is slow or docker is slow due to timing conditions). It's better than it was, though. Please take a look. Specifically, this line does not seem great https://github.com/heroku/buildpacks-ruby/pull/172/files#diff-17c88161984214fafac5afdc89a5919f2dbcfb1165f2560fd255809316095fe8R131. 

I applied all of your requested changes. Thanks for the 👀 . I've asked @Malax for another round of reviews.

buildpacks/ruby/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
edmorley, line 9: The https://keepachangelog.com/en/1.0.0/ changelog style has category H3 headings like `### Added` etc - but there aren't any used in this changelog at the moment?

buildpacks/ruby/tests/integration_test.rs
```diff
@@ -112,6 +111,51 @@ fn test_ruby_app_with_yarn_app() {
...
+            context.start_container(
+                ContainerConfig::new()
+                    .entrypoint("launcher")
+                    .envs(vec![
+                        ("DYNO", "web.1"),
+                        ("PORT", "1234"),
+                        ("AGENTMON_DEBUG", "1"),
+                        ("HEROKU_METRICS_URL", "example.com"),
+                    ])
+                    .command(["while true; do sleep 1; done"]),
```
edmorley, line 131: > I fixed the tests (https://github.com/heroku/buildpacks-ruby/pull/172#issuecomment-1714857517) to be more deterministic, but it's still based on a timeout (which can fail if the machine is slow or docker is slow due to timing conditions). It's better than it was, though. Please take a look. Specifically, this line does not seem great https://github.com/heroku/buildpacks-ruby/pull/172/files#diff-17c88161984214fafac5afdc89a5919f2dbcfb1165f2560fd255809316095fe8R131.

The `while started.elapsed() < Duration::from_secs(20)` timeout seems fine to me -- it's no different from some of the other timeouts we have in both buildpack implementations or integration tests IMO (eg download timeouts).

The `while true; do sleep 1; done` is fine too - though searching now I just learned that GNU `sleep` supports an `infinity` option too:
https://unix.stackexchange.com/questions/366077/is-there-a-linux-command-that-does-nothing-but-never-exits

That said, if you really didn't like the sleep, perhaps the testcase could be updated to have a `Procfile` that actually starts Puma, and then the testcase also has the advantage of exercising the "can barnes actually report to the statsd server successfully" part too. (It's fine if you want to skip this for now and handle later.)

edmorley (approved):
Thank you!

buildpacks/ruby/tests/integration_test.rs
schneems: I like `sleep infinity`. Functionally it's about the same, but it seems more intentional. However when I tried it I started to get failures (even after bumping my max time to 30 seconds), so it might not be interpreted as the same thing. 

```
thread 'test_barnes_app' panicked at 'assertion failed: `(left contains right)`
left (unescaped):
/layers/heroku_ruby/metrics_agent/agentmon_loop already running.
Logging enabled via `AGENTMON_DEBUG="1"`. To disable `unset AGENTMON_DEBUG`

left (escaped): `"/layers/heroku_ruby/metrics_agent/agentmon_loop already running.\nLogging enabled via `AGENTMON_DEBUG=\"1\"`. To disable `unset AGENTMON_DEBUG`"`
right: `"Booting agentmon_loop"`', buildpacks/ruby/tests/integration_test.rs:152:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We might want to call out: "Here's how to boot a container that will run until your block exits" in the docs for a use case like this.

buildpacks/ruby/CHANGELOG.md
schneems: I am not totally on board with the full keep-a-changelog (I love the project, but I don't use the full format in any of my projects). I prefer to have those annotations in line if they're required.

If we have a strong schema for our changelogs, we need to have validation on push to enforce them, or errors will build up over time.

I originally proposed using changelog annotations for buildpack release versioning, but AFAIK there are no plans for those annotations to be used in a meaningful way.

Malax: One reason to at least adhere to the format strictly for now is the fact that the buildpacks I saw do. So that we have at least consistency. We can challenge the status quo of course, but I don't think this PR is the right battlefield. :)

buildpacks/ruby/src/layers/metrics_agent_install.rs
```diff
@@ -0,0 +1,284 @@
...
+///
+/// - Repo: <https://github.com/heroku/agentmon>
+/// - Releases: <https://github.com/heroku/agentmon/releases>
+///
+/// To get the latest s3 url:
+///
+/// ```shell
+/// $ curl https://agentmon-releases.s3.us-east-1.amazonaws.com/latest
+/// ```
+const DOWNLOAD_URL: &str =
```
Malax, line 27: I would love to see at least some basic checksum validation after the download to improve our security posture.

Malax (changes requested):
Thanks for making the requested changes and sorry for the delay with the re-review! Apart from the checksum check, I see nothing blocking.

buildpacks/ruby/tests/integration_test.rs
edmorley: > We might want to call out: "Here's how to boot a container that will run until your block exits" in the docs for a use case like this.

I'm not sure that's really necessary, since:
- most libcnb-test integration tests should either (a) be running one-off commands using `run_shell_command` (as recommended already in the docs), or (b) if they need to be running servers, then they should actually run the servers not fake processes (ideally the tests here would be running Puma as mentioned above)
- there's a tipping point for docs-length over which people stop reading or miss important details because they are hidden in-between less important information, and so we should try not to add general programming/operating system knowledge in here unless it's going to affect lots of users and also hard to figure out (in this case the first result for `"linux command that never exits"` gives the exact answer)

buildpacks/ruby/src/layers/metrics_agent_install.rs
```diff
@@ -153,6 +158,15 @@ impl Layer for MetricsAgentInstall {
     }
 }
 
+// Check integrity of download, `sha_some` is awesome.
+fn sha_some(path: &Path) -> Result<String, std::io::Error> {
```
Malax, line 162: Can we just use https://github.com/heroku/libcnb.rs/blob/5e9aeb916d1706d81b90948913050f6589726c87/libherokubuildpack/src/digest.rs?

Malax (changes requested):

buildpacks/ruby/CHANGELOG.md
schneems: I updated it to use that format and also rev-d to use the latest version of the keep-a-changelog format: 1.1.0

buildpacks/ruby/src/layers/metrics_agent_install.rs
schneems: Are you asking me to make my code **less** `sha_some` :p 

Funny that it has the exact same signature. I didn't have `libherokubuildpack` pulled into this project, but I'll move to use the shared tooling 🎉 thanks for the recommendation!

buildpacks/metrics-agent/src/layers/install_agentmon.rs
schneems: Updated with a SHA 256 checksum using libherokubuildpack. Please take another look

schneems:
I've updated to use a checksum with the `libherokubuildpack` shared tooling.

Malax (approved):

PR #175, opened 2023-07-25
Update error messages

schneems:
Iteration based_ on this feedback doc: https://docs.google.com/document/d/1uixuWY91bRl4BX1eIdsPpkPAc3j5g6KYdGGYyCEf7lY/edit

[W-13815542](https://gus.lightning.force.com/a07EE00001WubQKYAZ)

You can see the updated output formatting live in the included CI tasks:

- print-pack-getting-started-output
- print-errors

PR #198, opened 2023-08-29
Update the logging interface

schneems:
Supercedes https://github.com/heroku/buildpacks-ruby/pull/175


Malax:
Do you want a thorough review on this or when you port this over to libherokubuildpack? Is that still the plan?

schneems:
> Do you want a thorough review on this or when you port this over to libherokubuildpack? Is that still the plan?

I think maybe give this a rough pass with an eye towards eventual integration, but I think we plan on a fairly exhaustive review on future `libherokubuildpack` work. Though if you both have the energy and feel the outcome could be better to really hammer both, then I say go for it. For the case of this specific PR the stakes are lower, so there's more room to experiment (saving that Colin is using my branch code so if we completely break an interface I want to at least warn him).

Malax:
@schneems in this case, I'd say we rubber stamp this one and do a proper review on the libherokubuildpack one only. This is a _huge_ PR and will take a lot of time to review and any change we make will break @colincasey.

schneems:
Sounds good. Let's move forward with that strategy.

commons/Cargo.toml
```diff
@@ -23,7 +27,12 @@ tempfile = "3"
 thiserror = "1"
 walkdir = "2"
 which_problem = "0.1"
+ascii_table = { version = "4.0.2", features = ["color_codes"] }
+const_format = "0.2.31"
```
edmorley, line 31: (For consistency with the dependency version strategy used already)

```suggestion
ascii_table = { version = "4", features = ["color_codes"] }
const_format = "0.2"
```

commons/Cargo.toml
```diff
@@ -23,7 +27,12 @@ tempfile = "3"
...
 
 [dev-dependencies]
+indoc = "2.0.1"
```
edmorley, line 34:
```suggestion
indoc = "2"
```

commons/Cargo.toml
```diff
@@ -23,7 +27,12 @@ tempfile = "3"
...
 filetime = "0.2"
 toml = "0.7"
+libcnb-test = "0.14.0"
+pretty_assertions = "1.4.0"
```
edmorley, line 38:
```suggestion
pretty_assertions = "1"
```

commons/Cargo.toml
```diff
@@ -23,7 +27,12 @@ tempfile = "3"
...
 walkdir = "2"
 which_problem = "0.1"
+ascii_table = { version = "4.0.2", features = ["color_codes"] }
+const_format = "0.2.31"
 
 [dev-dependencies]
+indoc = "2.0.1"
 filetime = "0.2"
 toml = "0.7"
+libcnb-test = "0.14.0"
```
edmorley, line 37: The `=` prefix here (used by the other libcnb related deps in this file) makes a difference - otherwise it's equivalent to `^0.14.0`

```suggestion
libcnb-test = "=0.14.0"
```

buildpacks/ruby/src/steps/detect_rake_tasks.rs
```diff
@@ -1,4 +1,9 @@
-use crate::build_output::{self, Section};
+#[allow(clippy::wildcard_imports)]
```
edmorley, line 1: (optional) I'm not really a fan of wildcard imports fwiw

buildpacks/ruby/src/gem_list.rs
```diff
@@ -1,5 +1,7 @@
-use crate::build_output::{RunCommand, Section};
-use commons::fun_run::{self, CmdMapExt};
+#[allow(clippy::wildcard_imports)]
+use commons::output::{fmt, section_log::*};
```
edmorley, line 2: (optional) I'm not really a fan of wildcard imports fwiw

buildpacks/ruby/src/gem_list.rs
```diff
@@ -1,5 +1,7 @@
...
+
```
edmorley, line 3: Remove the newlines to ensure all imports are sorted properly

edmorley (approved):
This is a very quick skim review of mainly the non-Rust parts, since at `+3,403 −1,507` this PR is too large to properly review - particularly since it mixes new APIs, error message content changes (as in the copy), CI changes etc.

buildpacks/ruby/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
```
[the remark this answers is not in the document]
schneems: I map these directly to semver and there's not a breaking change here (unless someone is relying on the logging output, which...maybe). As this is more mature I might count a change to the log output as a major bump, but I view this more like a bugfix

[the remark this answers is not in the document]
edmorley: I don't think that's the keepachangelog.com intended meaning behind "Changed" vs "Fixed" though.

> `Changed` for changes in existing functionality.
> ...
>  `Fixed` for any bug fixes. 

ie: The "Changed" definition doesn't say "breaking change" in its description. And "Fixed" explicitly calls them "bug fixes", and the logging change wasn't fixing a bug.

buildpacks/ruby/src/steps/detect_rake_tasks.rs
schneems: I removed all `section_log::*` imports, but left the `build_log::*` ones as those are needed for the API (the BuildLog struct implements several traits that must be imported for it to function correctly).

edmorley:
@schneems Is this ready for merge?

PR #223, opened 2023-10-23
Update buildpack metadata in buildpack.toml

edmorley:
Adjusts the buildpack `name`, `description` and `keywords` in `buildpack.toml` to match the style discussed in:
https://github.com/heroku/cnb-builder-images/issues/408

These fields are used by the CNB registry and can also be seen in the output of `pack builder inspect`. They are not used by `pack build` or Kodon.

In addition, I've cleaned up the `[metadata]` table/subtables to match the more concise style used by the Procfile and Python CNBs.

GUS-W-14121598.

### DIFF
--- a/buildpacks/ruby/CHANGELOG.md
+++ b/buildpacks/ruby/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


buildpacks/ruby/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
```
schneems, line 10:
```suggestion
### Fixed
```

schneems (approved):

PR #233, opened 2023-11-06
Bump fun_run

schneems:
Fixes double indenting the first line of streamed commands https://github.com/schneems/fun_run/pull/3

### DIFF
--- a/commons/Cargo.toml
+++ b/commons/Cargo.toml
@@ -28,7 +28,7 @@ thiserror = "1"
 walkdir = "2"
 ascii_table = { version = "4", features = ["color_codes"] }
 const_format = "0.2"
-fun_run = "0.1"
+fun_run = "0.1.1"
 
 [dev-dependencies]
 indoc = "2"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -457,9 +457,9 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
 name = "fun_run"
-version = "0.1.0"
+version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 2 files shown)


commons/Cargo.toml
```diff
@@ -28,7 +28,7 @@ thiserror = "1"
 walkdir = "2"
 ascii_table = { version = "4", features = ["color_codes"] }
 const_format = "0.2"
-fun_run = "0.1"
+fun_run = "0.1.1"
```
edmorley, line 31: The existing style used in this repo (and across all of our CNB repos) is to only put the major version in `Cargo.toml` and leave version resolution within that major version range to the lockfile.

edmorley (commented):
A quicker way to pick up updates like this (that saves having to create a PR manually or wait for review) is to trigger a manual Dependabot run.

To do this, go to "Insights" -> Dependency graph -> Dependabot -> click "Recent update jobs" for the Cargo job type, which lands you on this page:
https://github.com/heroku/buildpacks-ruby/network/updates/15241710/jobs

From there, press "Check for updates" - and Dependabot will start a new run :-)

commons/Cargo.toml
edmorley: @schneems I've triggered a Dependabot run manually now - which opened #234, that contains the `fun_run` update (amongst a couple of small other updates).

One thing I noticed that was the `fun_run` crate is missing some metadata (such as the GitHub repo URL) - which is why the changelog didn't show up on the Dependabot PR, and why there is no GitHub repo link on the crates.io page:
https://crates.io/crates/fun_run

For an example of the metadata you might want to include, see libcnb's manifest:
https://github.com/heroku/libcnb.rs/blob/5ab4b036a8dd18dc2d25a3d3ee49ed66ef3ec9d3/libcnb/Cargo.toml#L7-L12

The upstream docs have some more details on each of those fields:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-repository-field

PR #243, opened 2023-11-28
Report io error when detect cannot read a file.

schneems:
It's best practice to use `try_exists()` instead of `exists()` when checking if a file exists or not because if there's a problem with the file (such as a permissions issue) then the check might fail even though the file is on disk. This seems like not a big deal, but can be very confusing when telling users "we didn't do X because you are missing file Y" but when they debug they see file Y and don't realize that the real problem isn't "missing file" but it's the hidden error preventing `exists()` from returning a true.

This commit uses fs_err_try_exists() from fs-err library that was introduced in: https://github.com/andrewhickman/fs-err/pull/48.

There are a few other places where `exists` is still called, those are nested in areas that don't already return a Result so they're harder to swap.

### DIFF
--- a/buildpacks/ruby/src/user_errors.rs
+++ b/buildpacks/ruby/src/user_errors.rs
@@ -6,7 +6,7 @@ use commons::output::{
     fmt::{self, DEBUG_INFO},
 };
 
-use crate::RubyBuildpackError;
+use crate::{DetectError, RubyBuildpackError};
 use fun_run::{CmdError, CommandWithName};
 use indoc::formatdoc;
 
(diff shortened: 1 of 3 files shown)


buildpacks/ruby/src/user_errors.rs
```diff
@@ -45,6 +45,62 @@ fn log_our_error(mut log: Box<dyn StartedLogger>, error: RubyBuildpackError) {
...
+
+                Debug using the above information and try again.
+            "});
+        }
+        RubyBuildpackError::BuildpackDetectionError(DetectError::PackageJson(error)) => {
+            log.announce().error(&formatdoc! {"
+                Error: `package.json` found with error
+
+                The Ruby buildpack detected a package.json file but it is not readable
+                due to the following errors:
```
colincasey, line 65:
```suggestion
                The Ruby buildpack detected a package.json file but it is not readable \
                due to the following errors:
```

buildpacks/ruby/src/user_errors.rs
```diff
@@ -45,6 +45,62 @@ fn log_our_error(mut log: Box<dyn StartedLogger>, error: RubyBuildpackError) {
...
+
+                {error}
+
+                If your application does not need any node dependencies installed,
+                you may delete this file and try again.
```
colincasey, line 70:
```suggestion
                If your application does not need any node dependencies installed, \
                you may delete this file and try again.
```

colincasey (changes requested):

colincasey (approved):

PR #246, opened 2023-12-19
Introduce Magic Migrate

schneems:
The goal of magic migrate is to simplify versioned metadata storage.

## The problem

Cloud native build packs use toml to store metadata about a layer between builds. We commonly use this to store information like what ruby version was downloaded or a sha of some kind. On the next build we can look at this information to determine if some expensive process (like downloading a binary) can be skipped. Essentially we treat it like a cache key.

If we cannot load (deserialize) the old metadata into the currently requested structure then the default behavior is to clear the cache. This means that either the programmer must be careful to never make backwards incompatible changes to the metadata, or risk triggering a cache invalidation.

Now consider that we cannot guarantee that the cache was generated from the last version of the buildpack. Someone might deploy, then wait several years before deploying again.

The classic Ruby buildpack has this problem. It's cache is unversioned so if a mistake is made in the cache structure or contents in one version, then the fix must be hardcoded and checked on every subsequent deploy of every future version https://github.com/heroku/heroku-buildpack-ruby/blob/453b13983b638d68d9d65ab89d36a2fc18128e4a/lib/language_pack/ruby.rb#L1270-L1332.

## Introducing magic migrate

Magic migrate doesn't make these problem go away, instead it makes the problem easier to reason about. 

When the schema of the metadata changes, the programmer can introduce a new struct and tell rust how to migrate from one version to the next using either `From` or `TryFrom` (if fallible). Then they use the corresponding magic migrate trait to tell rust how to walk this chain backwards. Now when we try to load data from disk it will try to load the latest struc, if it can't it will go to the one before, and so-on. Once it finds the original serialized struct, it converts it forwards, one step at a time until we arrive at the currently desired struct.

Now instead of trying to hold all possible cache state in mind from all possible versions of the code, the programmer only needs to know how to make each conversion one by one.

### DIFF
--- a/commons/src/magic_migrate.rs
+++ b/commons/src/magic_migrate.rs
@@ -1,0 +1,382 @@
+use serde::de::DeserializeOwned;
+use std::any::{Any, TypeId};
+use std::fmt::Debug;
+
+/// Magically migrate serialized toml structs to the latest version
+///
+/// Approach:
+///
+/// If every struct knows the type of the one that came before it, and
+/// if every struct can `Into` the type that comes after it. Then,
+/// We build a type chain (using traits) to recurse backwards to find the
+/// first struct a given string can serialize into. Once we've found a
+/// serializable struct, we serialize it, then convert to the struct ahead
+/// of it via `into`.
+///
+/// This requires only **a tiny bit** of type inspection which is used to
+/// stop the recursion. The first struct in the chain has no predecessor
+/// it specifies `Before: Self`. When this is reached, we know that the
+/// type cannot be serialized to any struct so `None` is returned.
+///
+/// Example:
+///
+/// ```rust
+/// use commons::magic_migrate::{MigrateToml, MigrateDe};
+/// use serde::{Deserialize, Serialize};
+///
+/// // Given a struct that is serialized somwhere
+/// #[derive(Deserialize, Serialize, Debug)]
+/// struct ContactV1 {
+///     name: String,
+/// }
+///
+/// // Tell Rust that it can migrate to itself
+/// impl MigrateToml for ContactV1 {
+///     type From = Self;
+/// }
+///
+/// // Define the next version you of data you wish to use
+/// #[derive(Deserialize, Serialize, Debug)]
+/// struct ContactV2 {
+///     name: String,
+///     title: Option<String>,
+///     first_initial: String,
+/// }
+///
+/// // Tell rust how to convert from one to the other manually
+/// impl From<ContactV1> for ContactV2 {
+///     fn from(value: ContactV1) -> Self {
+///         ContactV2 {
+///             name: value.name.clone(),
+///             title: None,
+///             first_initial: value.name.chars().next().unwrap().to_string(),
+///         }
+///     }
+/// }
+///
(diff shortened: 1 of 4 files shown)


commons/src/magic_migrate.rs
```diff
@@ -0,0 +1,381 @@
...
+
+        let v3 = YoloV3::try_from_str_migrations(&toml_string).unwrap();
+        assert_eq!(
+            Err(YoloMigrationError::String(String::from(
+                "Cannot build a valid address from only a title"
+            ))),
+            v3
+        );
+    }
+}
```
schneems, line 381: Mentioning one known issue:

There's currently also an ABA problem with the cache. If you accidentally use the same struct field names, but with different logical semantics an older cache entry could serialize into a newer struct. Magic migrate does nothing to solve this problem. We should add recommendations in the future on best practices. One possible path forward is to use tagging (https://serde.rs/enum-representations.html#adjacently-tagged) to enforce that each struct's version is preserved as it's serialized:

```ruby
#[derive(Deserialize, Serialize, Debug, Clone, Eq, PartialEq)]
#[serde(tag = "v1")]
struct ContactV1 {
    name: String,
}

#[derive(Deserialize, Serialize, Debug, Clone, Eq, PartialEq)]
#[serde(tag = "v2")]
struct ContactV2 {
    name: String,
    title: Option<String>,
    first_initial: String,
}
```

However this could also introduce other issues, like if someone forgets to add a tag, or accidentally adds the wrong tag value, etc. In a perfect world we would also enforce that no tags in the chain are the same. Internally I don't know if we have access to that information.

schneems: I misunderstood how the `tag = ` works. It is the key and the name of the struct will be the value so the above would serialize to `v1 = \"ContactV1\"\nname = \"richard\""` so we would probably want to do something like:

```ruby
#[derive(Deserialize, Serialize, Debug, Clone, Eq, PartialEq)]
#[serde(tag = "ser_ver")]
struct ContactV1 {
    name: String,
}

#[derive(Deserialize, Serialize, Debug, Clone, Eq, PartialEq)]
#[serde(tag = "ser_ver")]
struct ContactV2 {
    name: String,
    title: Option<String>,
    first_initial: String,
}
```

Which would serialize to `"ser_ver = \"ContactV1\"\nname = \"richard\""`. I'm still not sure if we can detect/enforce that this feature is enabled.

schneems: I found `deny_unknown_fields` which is exactly what I want, but it seems to be incompatible with `tag`. I opened an issue https://github.com/serde-rs/serde/issues/2666

schneems: - Asking on stack overflow https://stackoverflow.com/questions/77700360/prevent-a-serialized-struct-in-rust-from-being-deserialized-into-a-different-one
- Asking on mastodon https://ruby.social/@Schneems/111619648760386358

schneems: I think this is out scope for this feature. It's up to developers to use MagicToml responsibly. If we develop best practices later we can annotate them via documentation.

PR #280, opened 2024-05-06
Support heroku-24

runesoerensen:
This PR adds support for heroku-24 including downloading arch specific Ruby binaries depending on the current host architecture.

The cache isn't currently invalidated if the host architecture changes.

### DIFF
--- a/buildpacks/ruby/src/layers/ruby_install_layer.rs
+++ b/buildpacks/ruby/src/layers/ruby_install_layer.rs
@@ -2,6 +2,7 @@ use commons::output::{
     fmt::{self},
     section_log::{log_step, log_step_timed, SectionLogger},
 };
+use libcnb::data::stack_id;
 
 use crate::{RubyBuildpack, RubyBuildpackError};
 use commons::gemfile_lock::ResolvedRubyVersion;
@@ -130,10 +131,18 @@ fn download_url(stack: &StackId, version: impl std::fmt::Display) -> Result<Url,
     let base = "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com";
     let mut url = Url::parse(base).map_err(RubyInstallError::UrlParseError)?;
 
-    url.path_segments_mut()
-        .map_err(|()| RubyInstallError::InvalidBaseUrl(String::from(base)))?
-        .push(stack)
-        .push(&filename);
+    if stack == &stack_id!("heroku-24") {
+        url.path_segments_mut()
+            .map_err(|()| RubyInstallError::InvalidBaseUrl(String::from(base)))?
+            .push(stack)
+            .push("amd64")
+            .push(&filename);
+    } else {
+        url.path_segments_mut()
+            .map_err(|()| RubyInstallError::InvalidBaseUrl(String::from(base)))?
+            .push(stack)
+            .push(&filename);
+    }
     Ok(url)
 }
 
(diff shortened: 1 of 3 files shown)


buildpacks/ruby/src/layers/ruby_install_layer.rs
```diff
@@ -130,10 +131,18 @@ fn download_url(stack: &StackId, version: impl std::fmt::Display) -> Result<Url,
...
 
-    url.path_segments_mut()
-        .map_err(|()| RubyInstallError::InvalidBaseUrl(String::from(base)))?
-        .push(stack)
-        .push(&filename);
+    if stack == &stack_id!("heroku-24") {
+        url.path_segments_mut()
+            .map_err(|()| RubyInstallError::InvalidBaseUrl(String::from(base)))?
+            .push(stack)
+            .push("amd64")
```
schneems, line 138: This is a great starting point. The only thing I would change is to use the architecture of the current machine. We're doing this in classic via shelling out https://github.com/heroku/heroku-buildpack-ruby/blob/fd6585411cbf86d363ad0824f53d8edba1d7afed/lib/language_pack/base.rb#L45-L57.

I'm hoping on a plane in a few hours and then out tomorrow, if you're available to take a stab at adding that extra logic that would be amazing 🙏🏻. If not, I can take this commit and add that on top.

I was working towards upgrading libcnb first, but seeing how compact this PR is, I think I like the idea of merging this in now and cutting a builder so people can play with it.

schneems (changes requested):
Need to support dynamically pulling based on currently running architecture (instead of only supporting amd64). I left a comment below. Thanks for the PR!

buildpacks/ruby/src/layers/ruby_install_layer.rs
runesoerensen: Yep shelling out is definitely not the best approach here.. This is somewhat related to https://github.com/heroku/buildpacks-nodejs/pull/814#discussion_r1580063265 -- e.g. whether to use the CNB targets, or the compile-time conditional (or `std::env::consts::ARCH` as in the node.js sample).

I alluded to a conversation I had with @Malax on this (re env provided target arch vs compile-time conditionals). What do you think about this @Malax (and did I remember the conversation correctly :-))?

runesoerensen: For sure, seems like it'd make a lot of sense to also use `context.target` in this buildpack when the buildpack is updated to newer `libcnb.rs` and `targets`.

I've updated this implementation to also factor in the current arch -- should be fairly straightforward to adapt this to use the `Target` instead.

buildpacks/ruby/src/layers/ruby_install_layer.rs
```diff
@@ -174,6 +194,9 @@ pub(crate) enum RubyInstallError {
     #[error("Invalid base url {0}")]
     InvalidBaseUrl(String),
 
+    #[error("Unsupported architecture: {0}")]
```
schneems, line 197: I think we want to go a little above and beyond and list out the supported architectures. 

In general, any time we're failing because a value was unexpected based on a list of currently known/supported values, I want to output both what was actually output (the current architecture) but also what was expected (either arm64 or amd64 etc.)

This can help the future development errors/issues for example if someone accidentally adds a space like `" arm64"`. Even if we think such an issue would be rare, I don't see a downside in adding that information to the error message (other than that requirement will influence the implementation, such as needing to move the information to a const, for example.

schneems (commented):
Thanks for iterating on this. I can take it over from here. I left one comment on an error message improvement as an FYI, but I don't need you to do anything. I'll grab you code to a new branch and iterate on some of Ed's comments.

schneems:
It made more sense to move over to a new branch than to try to replay main on top of this. I am doing the work over at https://github.com/heroku/buildpacks-ruby/pull/284. I want to be mindful to not delete this branch until `heroku/ruby` has heroku-24 support on main and a release as I know some might be using this branch from git.

edmorley:
> as I know some might be using this branch from git.

The buildpack can't be used directly from a branch since it would need compiling. The only other scenario is where someone might be using a shared crate, but this PR doesn't touch shared code.

schneems:
I've merged #284. I've queued up a major release GHA (since there's something marked as CHANGED due to the lifecycle requirement changing).

edmorley:
Resolved by #284.

PR #284, opened 2024-05-16
Support Heroku-24 and ARM architecture

schneems:
Heroku 24 stack image supports both arm64 and amd64. The Ruby binaries are available on heroku-24 compiled to these two stacks https://github.com/heroku/docker-heroku-ruby-builder/pull/38. The buildpack adds support by:

- Updating the `buildpackl.toml` to include ARM architecture
- Using architecture information to build the URL for Ubuntu 24.04 (heroku-24)
- Integration testing on all supported Ubuntu versions

### DIFF
--- a/buildpacks/ruby/CHANGELOG.md
+++ b/buildpacks/ruby/CHANGELOG.md
@@ -7,7 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- The buildpack now implements Buildpack API 0.10 instead of 0.9, and so requires `lifecycle` 0.17.x or newer. ([#283](https://github.com/heroku/buildpacks-ruby/pull/283/files#commit-suggestions))
+### Changed
+
+- The buildpack now implements Buildpack API 0.10 instead of 0.9, and so requires `lifecycle` 0.17.x or newer. ([#283](https://github.com/heroku/buildpacks-ruby/pull/283))
+
+### Added
+
+- Added support for Ubuntu 24.04 (and thus Heroku-24 / `heroku/builder:24`). ([#284](https://github.com/heroku/buildpacks-ruby/pull/284))
+
+### Fixed
 
 ## [2.1.3] - 2024-03-18
 
--- a/buildpacks/ruby/CHANGELOG.md
+++ b/buildpacks/ruby/CHANGELOG.md
@@ -15,8 +15,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added support for Ubuntu 24.04 (and thus Heroku-24 / `heroku/builder:24`). ([#284](https://github.com/heroku/buildpacks-ruby/pull/284))
 
-### Fixed
-
 ## [2.1.3] - 2024-03-18
 
 ### Changed
--- a/buildpacks/ruby/tests/integration_test.rs
+++ b/buildpacks/ruby/tests/integration_test.rs
@@ -289,9 +289,9 @@ fn amd_arm_builder_config(builder_name: &str, app_dir: &str) -> BuildConfig {
 
     match builder_name {
         "heroku/builder:24" if cfg!(target_arch = "aarch64") => {
-            builder.target_triple("aarch64-unknown-linux-musl")
+            config.target_triple("aarch64-unknown-linux-musl")
         }
-        _ => builder.target_triple("x86_64-unknown-linux-musl"),
+        _ => config.target_triple("x86_64-unknown-linux-musl"),
     };
     config
 }
--- a/buildpacks/ruby/tests/integration_test.rs
+++ b/buildpacks/ruby/tests/integration_test.rs
@@ -16,6 +16,11 @@ use ureq::Response;
 #[test]
 #[ignore = "integration test"]
 fn test_migrating_metadata() {
+    // This test is a placeholder for when a change modifies metadata structures.
+    // Remove the return and update the `buildpack-ruby` reference to the latest version.
+    #![allow(unreachable_code)]
+    return;
+
     let builder = "heroku/builder:22";
     let app_dir = "tests/fixtures/default_ruby";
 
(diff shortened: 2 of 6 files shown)


buildpacks/ruby/CHANGELOG.md
```diff
@@ -7,7 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- The buildpack now implements Buildpack API 0.10 instead of 0.9, and so requires `lifecycle` 0.17.x or newer. ([#283](https://github.com/heroku/buildpacks-ruby/pull/283/files#commit-suggestions))
+### Changed
+
+- The buildpack now implements Buildpack API 0.10 instead of 0.9, and so requires `lifecycle` 0.17.x or newer. ([#283](https://github.com/heroku/buildpacks-ruby/pull/283))
+
+### Added
+
+- Added support for Ubuntu 24.04 (and thus Heroku-24 / `heroku/builder:24`). ([#284](https://github.com/heroku/buildpacks-ruby/pull/284))
+
+### Fixed
```
edmorley, line 18:
```suggestion
```

buildpacks/ruby/CHANGELOG.md
```diff
@@ -7,7 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- The buildpack now implements Buildpack API 0.10 instead of 0.9, and so requires `lifecycle` 0.17.x or newer. ([#283](https://github.com/heroku/buildpacks-ruby/pull/283/files#commit-suggestions))
+### Changed
```
edmorley, line 10: Good spot. I missed this in the last PR review, doh.

buildpacks/ruby/tests/integration_test.rs
```diff
@@ -57,9 +62,26 @@ fn test_default_app_ubuntu20() {
...
+                r#"`BUNDLE_BIN="/layers/heroku_ruby/gems/bin" BUNDLE_CLEAN="1" BUNDLE_DEPLOYMENT="1" BUNDLE_GEMFILE="/workspace/Gemfile" BUNDLE_PATH="/layers/heroku_ruby/gems" BUNDLE_WITHOUT="development:test" bundle install`"#);
+
+            assert_contains!(context.pack_stdout, "Installing webrick");
+        },
+    );
+}
+
+#[test]
+#[ignore = "integration test"]
+fn test_default_app_latest_distro() {
```
edmorley, line 82: (Can be handled in a later PR) There is currently no testing of arm64 in CI. 

However, adding that would meaning needing to switch to a matrix style of testing, where the tests themselves all run on the input builder, and CI picks a different builder (and potentially architecture) per job in the matrix. For Python I make sure to run all python version related tests on all builders (so I know each compiled python archive works as expected, can find all dynamic libraries in that builder, hasn't been accidentally overwritten on S3 etc), and then on the main builder I run all tests.

As such, I'm fine with you making those CI changes in a later PR.

buildpacks/ruby/tests/integration_test.rs
```diff
@@ -57,9 +62,26 @@ fn test_default_app_ubuntu20() {
...
+    TestRunner::default().build(
+        BuildConfig::new("heroku/builder:24", "tests/fixtures/default_ruby"),
```
edmorley, line 84: This test fails for me locally, unless I add `.target_triple("aarch64-unknown-linux-musl")` because it's always compiling the buildpack for amd64 even if run on an arm64 host (and for `heroku/builder:24` if the host is arm64, then the arm64 image will be pulled by default).

Does this pass for you locally? If so, it might mean you have a global `DOCKER_DEFAULT_PLATFORM` set, which is masking the problem (I wouldn't set this globally in your shell).

To resolve the issue, you'll need something like:
https://github.com/heroku/buildpacks-python/blob/de378063c83e1eec6302e79f93f189162b5bbbbf/tests/mod.rs#L31-L38

(Having CI testing of arm64 would have caught this)

buildpacks/ruby/CHANGELOG.md
schneems: I wrote down an idea about adding a formatting check step and Colin seems supportive https://github.com/heroku/buildpacks/discussions/23.

buildpacks/ruby/CHANGELOG.md
schneems: I think that I would like these sections to already be in the CHANGELOG and then we just fill them out. That way you don't end up with two people adding a `Fixed` section in their PR and having extra conflicts to deal with. It's a bit more explicit as well, if someone is less familiar with keep a changelog format then they can tell at a glance that there were no "fixed" changes in this release. Not that useful now, but helpful for cases where there are no "changed" or "added" changes.

It could be more explicit with a `- None` bulletpoint under it, but that would lead to conflicts as it needs to be deleted.

I'm okay to take it out, but I'm suggesting we pre-populate this section when we cut a new release with the most common types of changes that we track.

buildpacks/ruby/tests/integration_test.rs
schneems: Updated to use similar logic https://github.com/heroku/buildpacks-ruby/pull/284. I verify that this test now passes for me locally.

I also opened https://github.com/heroku/libcnb.rs/issues/825.

buildpacks/ruby/CHANGELOG.md
edmorley: With that workflow, would you leave the empty sections in when you perform the actual release, or remove at that point? I don't like the idea of leaving empty sections in the actual release (given that there are 6 categories mentioned on https://keepachangelog.com/en/1.1.0/ and typically each release only has 1-2 categories used per release entry).

buildpacks/ruby/tests/integration_test.rs
```diff
@@ -257,3 +281,17 @@ fn frac_seconds(seconds: f64) -> Duration {
 }
 
 const TEST_PORT: u16 = 1234;
+
+// TODO: Once Pack build supports `--platform` and libcnb-test adjusted accordingly, change this
+// to allow configuring the target arch independently of the builder name (eg via env var).
+fn amd_arm_builder_config(builder_name: &str, app_dir: &str) -> BuildConfig {
+    let mut builder = BuildConfig::new(builder_name, app_dir);
```
edmorley, line 288:
```suggestion
    let mut config = BuildConfig::new(builder_name, app_dir);
```

buildpacks/ruby/tests/integration_test.rs
```diff
@@ -257,3 +281,17 @@ fn frac_seconds(seconds: f64) -> Duration {
...
+
+    match builder_name {
+        "heroku/builder:24" if cfg!(target_arch = "aarch64") => {
+            builder.target_triple("aarch64-unknown-linux-musl")
+        }
+        _ => builder.target_triple("x86_64-unknown-linux-musl"),
+    };
+    builder
```
edmorley, line 296:
```suggestion
    config
```

edmorley (approved):

buildpacks/ruby/CHANGELOG.md
schneems: I wrote a versioning RFC previously. I originally wanted the contents of the changelog to drive rev-ing the version.

I think we should strongly tie the tags we use in the CHANGELOG to versioning i.e. major, minor, patch. I would want those mapped to "keep a changelog" tags. Some are redundant or subsets. We can choose to only implement a subset of those tags by default and if we choose to extend later, then that's okay too.

PR #290, opened 2024-06-10
update builder to heroku-24 on print getting started guide github action

lilacstella:
Heroku 24 is now a trusted and suggested builder in [pack](https://github.com/buildpacks/pack/pull/2178). However, now running pack build with heroku-22 now fails due to a bug that suggests the builder is not trusted even though it is configured to be. 
```
Run pack build my-image --force-color --builder heroku/builder:22 --buildpack heroku/nodejs-engine --buildpack packaged/x86_64-unknown-linux-musl/debug/heroku_ruby --path tmp/ruby-getting-started --pull-policy never 
ERROR: failed to build: fetching lifecycle image: image buildpacksio/lifecycle:0.19.6 does not exist on the daemon: not found
``` 

We can update the version of the heroku builder that the github workflow for printing getting started is using in order for this action to pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Pull builder and run images
         run: |
           docker pull "heroku/builder:24"
-          docker pull "heroku/heroku:24-cnb"
+          docker pull "heroku/heroku:24"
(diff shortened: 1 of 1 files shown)


schneems (approved):
Thanks for getting this up and running!

PR #293, opened 2024-06-14
Test getting started guide with ARM && AMD

schneems:
Previously, all Heroku runners used AMD (x86) CPU architecture. We added the ability to build and run on an ARM machine (aarch64, like an M3 Mac), but building binaries for ARM required a beta GitHub Action (GHA) runner. The GHA runner beta required extra invocations to setup tools on the machine (such as installing docker). 

The GHA ARM runners are now out of beta. This PR adds a test that exercises the ARM codepath on CI in addition to the existing AMD codepath.

GUS-W-15805220

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,16 +60,20 @@ jobs:
         run: cargo test --locked -- --ignored
 
   print-pack-getting-started-output:
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.target == 'aarch64-unknown-linux-musl' && 'pub-hk-ubuntu-22.04-arm-medium' || 'ubuntu-latest' }}
+    strategy:
+      matrix:
+        target: ["aarch64-unknown-linux-musl", "x86_64-unknown-linux-musl"]
+      fail-fast: false
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Install musl-tools
-        run: sudo apt-get install musl-tools --no-install-recommends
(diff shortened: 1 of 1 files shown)


edmorley (commented):
Thank you for looking into ARM64 testing! 

Whilst this change does ensure that the buildpack build succeeds with an ARM64 platform (which confirms a few things, like: is the `buildpack.toml` `[[targets]]` metadata correct, and does the buildpack at least try and run on ARM vs failing with an error during the build), it doesn't actually confirm that the resultant app image is valid.

For example, if the Ruby CNB implementation forgot to check `context.target` during the build, and hardcoded an AMD64 path for some binary - the build could still succeed, but the app image then fail to run at runtime?

schneems:
> For example, if the Ruby CNB implementation forgot to check context.target during the build, and hardcoded an AMD64 path for some binary - the build could still succeed, but the app image then fail to run at runtime?

The ruby binary is exercised several times in the buildpack on this app via commands. The commands `gem install bundler`, `gem list` and `bundle exec rake assets:precompile` all invoke the Ruby interpreter. They will fail if they're the wrong version or the wrong architecture as bundler ensures the Ruby version matches the `Gemfile.lock`. 

I don't think we need the added complexity of adding an extra dimension to the matrix. Or rather, I'm wary of it.

edmorley:
Makes sense - happy to land this for now - though longer term I still think there is some benefit of testing a subset of the `libcnb-test` integration tests on multiple builder image/arch variants :-)

edmorley (approved):

PR #297, opened 2024-06-17
add node js installation clarifications to new ruby buildpack

### LINKED ISSUE #296: Lacking documentation on change in javascript runtime installation behavior between classic and current ruby buildpack
The ruby buildpack failed to automatically build in JRuby in https://github.com/heroku/jruby-getting-started/issues/26. This is due to the lack of a javascript runtime when building without the Node.js buildpack, there should be documentation regarding that when users upgrade from the classic version to CNBs. More details in #295.

We should add some clarification in https://github.com/heroku/buildpacks-ruby/blob/main/docs/upgrading.md.

lilacstella:
Adds a clarification regarding how users will need to structure their repository in order for the Node.js runtime to automatically be included as part of the build process. 

Resolves #296


docs/upgrading.md
```diff
@@ -9,6 +9,7 @@ This buildpack does not port all behaviors of the [classic buildpack for Ruby ](
...
 - Default Ruby versions are no longer sticky. Before bundler started recording the Ruby version in the `Gemfile.lock` it was common for developers to forget to declare their Ruby version in their project. When that happens they would receive the default Ruby version from the classic buildpack. To guard against instability the classic buildpack would record that version and use it again on all future deploys where no explicit version was specified. Now it's less likely for an app to deploy without a ruby version in their `Gemfile.lock`, and we do not want to encourage relying on a default for stability (as this otherwise breaks dev-prod parity).
+- The Node.js runtime will only be installed if there is a package.json file present at the root of the repository. A default Node.js version will no longer be installed based on the presence of gems that imply the dependency of a javascript runtime. Previously, the classic buildpack would install a version of Node.js if [a gem such as execjs or webpacker was detected](https://github.com/heroku/heroku-buildpack-ruby/blob/168cb5d556f154e4eba57c797d04b18a4d8b354c/lib/language_pack/ruby.rb#L944-L946).
```
schneems, line 12:
```suggestion
- The Node.js runtime will only be installed if there is a `package.json` file present at the root of the repository. A default Node.js version will no longer be installed based on the presence of gems that imply the dependency of a javascript runtime. Previously, the classic buildpack would install a version of Node.js if [a gem such as execjs or webpacker was detected](https://github.com/heroku/heroku-buildpack-ruby/blob/168cb5d556f154e4eba57c797d04b18a4d8b354c/lib/language_pack/ruby.rb#L944-L946).
```

schneems (approved):
Thanks!

PR #320, opened 2024-09-09
Migrate to the Struct API and output via `bullet_stream`

schneems:
This PR deprecates the `commons::build_output` module in favor of the bullet stream crate. It also uses the updated "struct API" from libcnb.rs. To minimize the change, I tried to avoid refactoring code as much as possible.

## Future refactors (not implemented)

I would like to eventually move the `layers/*.rs` files out and get rid of that nested directory. It adds nothing but overhead. I would also like to refactor the enum based approach to layer invalidation. It's overly complex and adds little value. Ideally I would like a list of all keys that changed and I experimented with a way to do this automatically via relying on toml comparisons in https://github.com/heroku/buildpacks-ruby/pull/203. I think this can be extracted and generalized.


buildpacks/ruby/src/layers/metrics_agent_install.rs
```diff
@@ -64,97 +60,53 @@ pub(crate) enum MetricsAgentInstallError {
...
+    )?;
+
+    let bin_dir = layer.path().join("bin");
+
+    match layer.state {
+        LayerState::Restored { .. } => {
+            bullet = bullet.sub_bullet("Using cached metrics agent");
         }
-    }
+        LayerState::Empty { .. } => {
```
runesoerensen, line 93: This doesn't handle the cases when the metrics agent is updated due to either another metrics agent URL, or invalid metadata

runesoerensen (commented):
This PR would be faster to review if it introduced the changes in at least 2 PRs instead:

1. Migrate to struct layer API.
2. Migrate to bullet_stream.
3. Handles deprecations in the commons/publicly shared crate (although this could also be done in 2, as it doesn't really have an impact on the review complexity like the other changes do).

PR #325, opened 2024-09-23
Update Ruby Install Layer to Struct API and bullet stream output

schneems:
Commits should be stand-alone.

### DIFF
--- a/buildpacks/ruby/src/layers/ruby_install_layer.rs
+++ b/buildpacks/ruby/src/layers/ruby_install_layer.rs
@@ -65,7 +65,7 @@ pub(crate) fn handle(
                 if let Some(diff) = metadata_diff(old, &metadata) {
                     (
                         RestoredLayerAction::DeleteLayer,
-                        SentenceList::new(&diff).to_string(),
+                        format!("due to change(s): {}", SentenceList::new(&diff)),
                     )
                 } else {
                     (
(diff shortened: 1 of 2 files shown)


buildpacks/ruby/src/layers/ruby_install_layer.rs
```diff
@@ -10,51 +27,107 @@ use crate::{
...
+                        RestoredLayerAction::DeleteLayer,
+                        format!("due to change(s): {}", SentenceList::new(&diff)),
+                    )
+                } else {
+                    (
+                        RestoredLayerAction::KeepLayer,
+                        "using cached version".to_string(),
+                    )
+                }
+            },
```
runesoerensen, line 76: Would it perhaps make sense to change `metadata_diff` to return a `Vec<String>`, and then check if it's empty here instead? Wrapping a vector in `Some` might be a bit redundant/unnecessary.

Just an idea, feel free to disregard!

```suggestion
            restored_layer_action: &|old: &Metadata, _| {
                let diff = metadata_diff(old, &metadata);
                if diff.is_empty() {
                    (
                        RestoredLayerAction::DeleteLayer,
                        format!("due to change(s): {}", SentenceList::new(&diff)),
                    )
                } else {
                    (
                        RestoredLayerAction::KeepLayer,
                        "using cached version".to_string(),
                    )
                }
            },
```

runesoerensen (approved):
Looks great! I left one comment, but feel free to merge as-is.

buildpacks/ruby/src/layers/ruby_install_layer.rs
schneems: FWIW https://github.com/heroku/buildpacks-ruby/blob/6f9ec54b51455fb5f3ff4e5c23768dae285e3855/commons/src/display.rs#L60 displays an empty iterator as "empty" by default so it would show `Clearing cache due to change(s): empty`.

I'll give the unwrapped Vec a try. 

Ultimately I want a declarative interface that could give me both consistency and reduce boilerplate. I've experimented with using the fact that metadata is serializable which means we have access to (some) reflection capabilities at runtime. I'm imagining something like this:

Adding annotations:

```rust
#[derive(MetadataDiff)]
pub(crate) struct MetadataV2 {
    pub(crate) distro_name: String,
    pub(crate) distro_version: String,

    #[diff(name: "CPU architecture")]
    pub(crate) cpu_architecture: String,

    #[diff(name: "Ruby version")]
    pub(crate) ruby_version: ResolvedRubyVersion,

    #[diff(ignore)]
    pub(crate) force_bundle_install_key: String,

    #[diff(ignore)]
    pub(crate) digest: MetadataDigest
}
```

Then a function could take in anything that implements that trait

```rust
fn metadata_diff<T: MetadataDiff>(old: T, now: T) -> DiffIter
```

I'm not planning on shipping that soon, but I wanted to mention where I want to land eventually.

schneems: I took your suggestion.

Also, despite saying I wouldn't implement the above soon, I realized I could do most of this without a macro (I just need one trait). This allows me to standardize the output in preparation for #326. It also allows me to test cache invalidation logic as a unit test without needing a full integration test.

schneems: I said so in slack, will repeat here. I've moved those changes to another PR.

PR #327, opened 2024-09-25
Shared layer logic for consistent output and unit testing

schneems:
PR #326 pointed out inconsistencies in the style of output for different layers and asked for additional tests. This refactor makes that request easier by introducing shared logic for invalid metadata and cache invalidation.

It's a ducktyping approach based on two traits:

- `magic_migrate::TryMigrate` provides an interface to (possibly) migrate invalid metadata or display errors
- `MetadataDiff` provides an interface to allow standardize cache invalidation and restoration logic

These allow me to remove the bulk of cache invalidation logic into a central shared location. They also expose a clean interface for unit testing cache invalidation logic, which can then be used on other layers.

I described some of the background desires in #325, comment: https://github.com/heroku/buildpacks-ruby/pull/325#discussion_r1773558240.

### DIFF
--- a/buildpacks/ruby/src/layers/ruby_install_layer.rs
+++ b/buildpacks/ruby/src/layers/ruby_install_layer.rs
@@ -327,7 +327,7 @@ version = "3.1.3"
     }
 
     #[test]
-    fn test_ruby_version_difference() {
+    fn test_ruby_version_difference_clears_cache() {
         let old = Metadata {
             ruby_version: ResolvedRubyVersion("2.7.2".to_string()),
             distro_name: "ubuntu".to_string(),
@@ -335,6 +335,12 @@ version = "3.1.3"
             cpu_architecture: "x86_64".to_string(),
         };
         let differences = old.diff(&old);
+        let actual = restored_layer_action(&old, &old);
+        assert!(matches!(
+            actual,
+            (libcnb::layer::RestoredLayerAction::KeepLayer, _)
+        ));
+
         assert_eq!(differences, Vec::<String>::new());
 
         let now = Metadata {
@@ -342,14 +348,10 @@ version = "3.1.3"
             ..old.clone()
         };
 
-        let differences = now.diff(&old);
-        assert_eq!(
-            differences,
-            vec![format!(
-                "Ruby version ({old} to {now})",
-                old = style::value("2.7.2"),
-                now = style::value("3.0.0")
-            )]
-        );
+        let actual = restored_layer_action(&old, &now);
+        assert!(matches!(
+            actual,
+            (libcnb::layer::RestoredLayerAction::DeleteLayer, _)
+        ));
     }
 }
--- a/buildpacks/ruby/src/layers/ruby_install_layer.rs
+++ b/buildpacks/ruby/src/layers/ruby_install_layer.rs
@@ -325,4 +325,31 @@ version = "3.1.3"
             "https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-22/ruby-2.7.4.tgz",
         );
     }
+
+    #[test]
+    fn test_ruby_version_difference() {
+        let old = Metadata {
+            ruby_version: ResolvedRubyVersion("2.7.2".to_string()),
+            distro_name: "ubuntu".to_string(),
+            distro_version: "20.04".to_string(),
+            cpu_architecture: "x86_64".to_string(),
+        };
+        let differences = old.diff(&old);
+        assert_eq!(differences, Vec::<String>::new());
+
+        let now = Metadata {
+            ruby_version: ResolvedRubyVersion("3.0.0".to_string()),
+            ..old.clone()
+        };
+
+        let differences = now.diff(&old);
+        assert_eq!(
+            differences,
+            vec![format!(
+                "Ruby version ({old} to {now})",
+                old = style::value("2.7.2"),
+                now = style::value("3.0.0")
+            )]
+        );
+    }
 }
--- a/buildpacks/ruby/src/layers/ruby_install_layer.rs
+++ b/buildpacks/ruby/src/layers/ruby_install_layer.rs
@@ -249,7 +249,7 @@ pub(crate) enum RubyInstallError {
 
 #[cfg(test)]
 mod tests {
-    use crate::layers::shared::restored_layer_action;
+    use crate::layers::shared::temp_build_context;
 
     use super::*;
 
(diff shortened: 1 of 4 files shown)


buildpacks/ruby/src/layers/ruby_install_layer.rs
```diff
@@ -362,4 +322,33 @@ version = "3.1.3"
...
+
+    #[test]
+    fn test_ruby_version_difference_clears_cache() {
+        let old = Metadata {
+            ruby_version: ResolvedRubyVersion("2.7.2".to_string()),
+            distro_name: "ubuntu".to_string(),
+            distro_version: "20.04".to_string(),
+            cpu_architecture: "x86_64".to_string(),
+        };
+        let differences = old.diff(&old);
```
colincasey, line 334: I think this test combines two distinct things that should be validated separately:
- The `diff` behavior should have it's own set of tests for the `MetadataDiff` trait implementation
- The `restored_layer_action` should have it's own set of of tests for cache layer action + expected output format

schneems: > The restored_layer_action should have it's own set of of tests for cache layer action + expected output format

I think that's a good idea. I'm trying to balance showing "just enough" of the functionality with a specific use case and was worried that a ton of tests make the PR line numbers seem larger than they really are. Via your comment it sounds like you're onboard with the approach and think it's worth it to invest more in stabilization and guarantees. If so, I agree.

schneems: > The diff behavior should have it's own set of tests for the MetadataDiff trait implementation

I'm not totally following here. Are you saying we should also test `MetadataDiff` in the shared module? If so: to test `restored_layer_action` in isolation I'll need to exercise the MetadataDiff trait. I think what you're wanting will fall out naturally, but the logic in those tests will be inherently coupled for now (which is okay, because they're two concepts that are coupled).

An alternative read could be that you feel the ruby layer module shouldn't have knowledge of the `restored_layer_action`.

If you meant something else, let me know.

colincasey: > An alternative read is that you're wanting to not call restored_layer_action from these tests and instead only test the MetadataDiff logic here and trust that restored_layer_action will do the right thing which is what I started out with.

That one. I don't mind seeing more unit test code. It's nice to see how various permutations and edge cases play out. 

Also, I like how you've structured the code so that it's possible to isolate out these relevant parts 😄

schneems: That makes sense. I view vaguely feel that testing only MetadataDiff as relying too much on the implementation. If change that Trait in some way that doesn't modify `diff` behavior but makes `restored_layer_action` not rely on it, then we're not asserting the end behavior we want. Unfortunately that desire to "test the behavior and not the implementation" is a bit of a rabbit hole.

I spent, perhaps too much time thinking on how to best structure this so that we rely on minimal coupling. It's a bit verbose so I put it in it's own gist https://gist.github.com/schneems/a4b7074793e8fdc1074c16f880c178f7 but if you want to chat about it, let's do that here. Interestingly enough that led me back to one of my first closed PRs against libcnb which was the desire to include some kind of `BuildContext` test generator.

schneems: I introduced a shorthand way to make a BuildContext that lets me exercise the MetadataDiff logic. https://github.com/heroku/buildpacks-ruby/pull/327/commits/36d14eb820ee2c1d6207173bb1c0d23fc0cfb3f4

I'll need to use another approach for the invalid metadata. Working on that now.

schneems: Ready for re-review. I added tests to shared and changed the interface I'm testing in the `ruby_install_layer`. 

The PR size ballooned, but 209 of the 397 lines are tests and helpers.

Please take a look. Also, @runesoerensen, this is ready for review if you can.

colincasey (changes requested):
@schneems I'm good with most of this code but the inability to choose between `build: true` and `launch: true` in the `cached_layer_write_metadata` function seems like it should either be re-evaluated or explained.

buildpacks/ruby/src/layers/shared.rs
```diff
@@ -0,0 +1,307 @@
...
+where
+    B: libcnb::Buildpack,
+    M: MetadataDiff + magic_migrate::TryMigrate + serde::ser::Serialize + std::fmt::Debug,
+    <M as magic_migrate::TryMigrate>::Error: std::fmt::Display,
+{
+    let layer_ref = context.cached_layer(
+        layer_name,
+        CachedLayerDefinition {
+            build: true,
+            launch: true,
```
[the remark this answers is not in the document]
schneems: I grep-d my codebase and I use `build: false` or `launch: false` in zero places. I see this as an opinionated default, but if you don't like the defaults, then you can drop down to using the individual components. That was my original idea anyway.

For a productionized/generic version I like the idea of a builder with defaults. For this PR, the current code meet my current needs.

buildpacks/ruby/src/layers/ruby_install_layer.rs
```diff
@@ -170,44 +126,46 @@ impl TryFrom<MetadataV1> for MetadataV2 {
...
+            ruby_version,
+        } = old;
+        if ruby_version != &self.ruby_version {
+            differences.push(format!(
+                "Ruby version ({old} to {now})",
+                old = style::value(ruby_version.to_string()),
+                now = style::value(self.ruby_version.to_string())
+            ));
+        }
+        if distro_name != &self.distro_name {
```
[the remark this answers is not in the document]
schneems: I had considered that previously. I like it. I'll run with it.

buildpacks/ruby/src/layers/shared.rs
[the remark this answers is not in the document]
colincasey: 👍 Just wanted to confirm this wasn't an oversight and you'd thought about this.

colincasey (approved):

buildpacks/ruby/src/layers/ruby_install_layer.rs
schneems: To bring it back, i ended up implementing the ability to call the top level function that's also being used in the layer which requires a BuildContext. Which is overkill, but I needed it anyway to be able to write tests for the shared logic.

I also added (just now) tests for the format of the MetadataDiff. I think in the future, we can rely on that interface for testing and leave a comment mentioning that the logic doubles up as cache invalidation logic.

Thanks for the review and suggestions!

buildpacks/ruby/src/layers/ruby_install_layer.rs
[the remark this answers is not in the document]
schneems: Resolving as I did it.

PR #330, opened 2024-10-03
Move bundle download to struct API and bullet stream

schneems:
Details in commits. The handle layer caching logic is exactly the same as the Ruby install layer https://github.com/heroku/buildpacks-ruby/blob/ec57995644a5feb3a89558918551373dc8d3541b/buildpacks/ruby/src/layers/ruby_install_layer.rs#L40-L58

### DIFF
--- a/buildpacks/ruby/src/layers/bundle_download_layer.rs
+++ b/buildpacks/ruby/src/layers/bundle_download_layer.rs
@@ -18,13 +18,27 @@ use libcnb::data::layer_content_metadata::LayerTypes;
 use libcnb::layer::{ExistingLayerStrategy, Layer, LayerData, LayerResult, LayerResultBuilder};
 use libcnb::layer_env::{LayerEnv, ModificationBehavior, Scope};
 use libcnb::Env;
-use serde::{Deserialize, Serialize};
+use magic_migrate::{try_migrate_deserializer_chain, TryMigrate};
+use serde::{Deserialize, Deserializer, Serialize};
+use std::convert::Infallible;
 use std::path::Path;
 use std::process::Command;
 
+pub(crate) type Metadata = MetadataV1;
+try_migrate_deserializer_chain!(
+    deserializer: toml::Deserializer::new,
+    error: MetadataError,
+    chain: [MetadataV1],
+);
+
 #[derive(Deserialize, Serialize, Debug, Clone)]
-pub(crate) struct Metadata {
+pub(crate) struct MetadataV1 {
     pub(crate) version: ResolvedBundlerVersion,
+}
+
+#[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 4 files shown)


buildpacks/ruby/src/layers/bundle_download_layer.rs
```diff
@@ -1,161 +1,163 @@
...
-                        Scope::All,
-                        ModificationBehavior::Prepend,
-                        "GEM_PATH", // Bundler is a gem too, allow it to be required
-                        gem_path,
-                    ),
-            )
-            .build()
-    }
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum MetadataError {
```
colincasey, line 80: So an error might happen during migration between metadata versions and you have this enum to represent that. There's currently only a single metadata version at the moment (`MetadataV1`) which implies that there are no migrations to perform so there are zero errors to represent. Therefore this empty enum exists more as a placeholder for future migrations?

buildpacks/ruby/src/layers/bundle_download_layer.rs
```diff
@@ -1,161 +1,163 @@
...
+    cmd.args(["--version", &metadata.version.to_string()]) // Specify exact version to install
+        .env_clear()
+        .envs(env);
 
-                Ok(ExistingLayerStrategy::Recreate)
-            }
-        }
-    }
-}
+    let short_name = fun_run::display(&mut cmd); // Format `gem install --version <version>` without other content for display
```
colincasey, line 99: interesting way to hide subsequent args but why not just show those extra args to the user?

buildpacks/ruby/src/layers/bundle_download_layer.rs
schneems: That's correct. I'm effectively laying out an empty structure that people can use later. It means that all layers that use these shared components will all be standardized. If you know how to modify one, you know how to modify all of them.

Technically I could use https://doc.rust-lang.org/std/convert/enum.Infallible.html instead. But that would require the first dev who needs to introduce a metadata migration to understand more about what's going on, enough to create a new enum and replace the error type.

buildpacks/ruby/src/layers/bundle_download_layer.rs
schneems: When presented with a failed build, the average user might get distracted by those extra arguments when they should be looking somewhere else. If the command fails, it prints out all the arguments, which is when a user might actually want to see them. Effectively: I see shelling out to this command as an implementation detail. I considered hiding it all together, but I think showing the short version peels back a little of the magic without being overwhelming.

Versus with `bundle install` that's a system-critical load-bearing source of support tickets and I want developers to be able to copy and paste the EXACT thing we run into their local machine.

PR #331, opened 2024-10-04
Schneems/refactor bundle install layer

schneems:


schneems:
I think i'm going to do this one again. I ended up effectively trying a few ideas and then going back to an earlier/easier one regarding how to get data out of the layer.

schneems:
Replaced by #332

PR #334, opened 2024-10-09
Continue replacing Layer trait with Struct API and using bullet stream

schneems:
Commits are small and descriptive.


runesoerensen (changes requested):
Should we perhaps add a changelog entry noting the `commons/src/cache/app_cache_collection.rs` deprecations?

Issue #302, opened 2024-07-15
Investigate FIFO asset cache cleaner with Pack creation time behavior


schneems:
Noted in https://github.com/heroku/buildpacks/discussions/22, `pack` mucks with creation time.

The Ruby CNB supports caching assets from `rake assets:precompile`, but we need a way to expire them. Beyond asset performance, it's also used for features such as sprocket's "last 3 versions" of an asset being made available. (In the asset pipeline when you modify a file, it generates a new output with a new SHA in the filename. But things like emails might reference older asset SHAs so you want to keep the old ones around for a little while, this is accomplished by storing assets generated in the cache between builds.)

Likewise, we need some way to invalidate cache entries or the cache size would grow indefinitely. The classic buildpack caps the memory limit of the asset directory and when it hits that limit, it deletes the oldest files first as newer files are more likely to still be used or referenced.

The problem (as I'm guessing) is that if `pack` is messing with creation time then it's effectively giving all files a very similar creation time even if they were created months/years apart. This could mean that the most recently generated assets are now cleared from cache instead of the oldest ones which will result in random asset lookup failures at runtime that will be difficult to diagnose and debug.

## TODO

- [x] Validate this is actually a problem.
- [ ] Investigate if a workaround is possible.
- [ ] Work with upstream to investigate long-term fixes.

schneems:
Comment is a work in progress. I'm trying to de-tangle all the information about build reproducibility

----

mtime

## Application use of mtime

Libraries and applications use the `mtime` of a file to determine last modified time of files. This is used in logic such as cache eviction of older files. While application owners and library authors might know that there are caveats around mtime information, it's generally considered to be a stable API to rely on.

## CNB project mtime implications

The cloud native buildpack project generates OCI images. One desired property of an OCI image is deterministic builds. That is, given the same inputs, the project produces the same image byte-for-byte. Timestamps complicate determinisim.

Another important feature of CNBs is rebasing https://buildpacks.io/docs/for-platform-operators/how-to/integrate-ci/pack/cli/pack_rebase/. Rebasing allows for updating system dependencies (like a version of curl, for example) without the need to completely re-build the entire image.

## Features

- By default images produce by pack have the same creation date of January 1, 1980. This is intentional so that the same source code input into the same builder will output the same image checksum. See https://buildpacks.io/docs/for-app-developers/concepts/reproducibility/

> For image creation commands (builder create, buildpack package, build) pack creates container images in a reproducible fashion.

[...]

[...]

PR #335, opened 2024-10-15
Test file copying preserves mtime

schneems:
Adds a test to ensure that a file copied via the AppCache feature preserves the file's original mtime. 

In #302 I determined that as long as the files copied into the cache dir preserve the original mtime then the FIFO cache cleaning logic that we rely on is valid. This test closes out this comment that asserted we need to verify whether or not the buildpack preserved mtime when copying files https://github.com/heroku/buildpacks-ruby/issues/302#issuecomment-2415128173.

#close 302

----
Commits are in historical order. I'm interested in feedback regarding tests, possibly some edge cases I've overlooked regarding test cases. app_cache.rs is approx 213 lines without comments or tests.

### DIFF
--- a/commons/src/cache/app_cache.rs
+++ b/commons/src/cache/app_cache.rs
@@ -448,4 +448,35 @@ mod tests {
         assert!(store.cache.join("lol.txt").exists());
         assert!(!store.path.join("lol.txt").exists());
     }
+
+    #[test]
+    fn mtime_preserved() {
+        let tmpdir = tempfile::tempdir().unwrap();
+        let cache_path = tmpdir.path().join("cache");
+        let app_path = tmpdir.path().join("app");
+
+        fs_err::create_dir_all(&cache_path).unwrap();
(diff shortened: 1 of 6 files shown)


commons/src/cache/app_cache.rs
```diff
@@ -448,4 +458,81 @@ mod tests {
...
+        let metadata = fs_err::metadata(store.cache.join("lol.txt")).unwrap();
+        let actual = filetime::FileTime::from_last_modification_time(&metadata);
+        assert_eq!(mtime, actual);
+
+        fs_err::remove_dir_all(&app_path).unwrap();
+        assert!(!store.path.join("lol.txt").exists());
+
+        store.load().unwrap();
+
+        let metadata = fs_err::metadata(store.cache.join("lol.txt")).unwrap();
```
schneems, line 534: Should be store.path and not store.cache

schneems:
Cleaned up and re-opened https://github.com/heroku/buildpacks-ruby/pull/336

PR #336, opened 2024-10-16
Close #302 Preserve mtime when copying to/from the cache

### LINKED ISSUE #302: Investigate FIFO asset cache cleaner with Pack creation time behavior

schneems:
Commits have more details. The short version is that I confirmed in https://github.com/heroku/buildpacks-ruby/issues/302#issuecomment-2415128173 that mtime behavior is preserved with the CNB cache. However, the fs_extra method of copying to/from the cache did not preserve mtime. The `cp_r` library is built to handle this case.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,6 +281,7 @@ dependencies = [
  "ascii_table",
  "byte-unit",
  "const_format",
+ "cp_r",
  "fancy-regex",
  "filetime",
  "fs-err",
@@ -320,6 +321,15 @@ dependencies = [
  "proc-macro2",
  "quote",
  "unicode-xid",
+]
+
+[[package]]
+name = "cp_r"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "837ca07dfd27a2663ac7c4701bb35856b534c2a61dd47af06ccf65d3bec79ebc"
+dependencies = [
+ "filetime",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,7 +285,6 @@ dependencies = [
  "fancy-regex",
  "filetime",
  "fs-err",
- "fs_extra",
  "fun_run",
  "glob",
  "indoc",
--- a/commons/CHANGELOG.md
+++ b/commons/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog for commons features
 
+## 2024-08-16
+
+### Fixed
+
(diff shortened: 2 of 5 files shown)


schneems:
Error on second run with the cache:

```
- Debug info
  - Could not move files out of the application to the cache.
    From: /workspace/tmp/cache/assets To: /layers/heroku_ruby/cache_tmp_cache_assets
    Error: creating directory: /layers/heroku_ruby/cache_tmp_cache_assets/sprockets: File exists (os error 17)

! Error caching frontend assets
!
! An error occurred while attempting to cache frontend assets, and the Ruby buildpack
! cannot continue.
!
! Ensure that the permissions on the files in your application directory are correct and that
! all symlinks correctly resolve.

ERROR: failed to build: exit status 1
ERROR: failed to build: executing lifecycle: failed with status code: 51
```

Is due to `cp_r`. I misread the docs it says:

> Overwrite existing directories or files.

However, it's under the banner "Missing features that could be added". It seems easy enough to implement following fs_err's API https://docs.rs/fs_extra/1.3.0/fs_extra/dir/struct.CopyOptions.html. 

Alternatively we could have an `on_conflict` API that took a `Fn(&Path, &Path) -> Result<&Path, Error>` where it passes in the path to both the from and to paths and lets you return the path to the one you want to keep (or optionally raise an error. It's a little more verbose, but would allow for all possible combinations with a single API.

schneems:
Closing in favor of #337

PR #337, opened 2024-10-17
Close #302 Copy Mtimes in the AppCache

### LINKED ISSUE #302: Investigate FIFO asset cache cleaner with Pack creation time behavior

schneems:
Each commit compiles and is intended to be small and readable. More information on why this option was chosen is in https://github.com/heroku/buildpacks-ruby/issues/302#issuecomment-2419661964

### DIFF
--- a/commons/CHANGELOG.md
+++ b/commons/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog for commons features
 
+## 2024-08-16
+
+### Fixed
+
+- `AppCache` will now preserve mtime of files when copying them to/from the cache (https://github.com/heroku/buildpacks-ruby/pull/336)
+
 ## 2024-08-15
 
 ### Changed
--- a/commons/Cargo.toml
+++ b/commons/Cargo.toml
@@ -32,6 +32,7 @@ sha2 = "0.10"
 tempfile = "3"
 thiserror = "1"
 walkdir = "2"
+filetime = "0.2"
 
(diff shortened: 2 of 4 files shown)


Malax (approved):
I didn't give this a super deep review, but I looked into it. It looks good to me! :)

Issue #333, opened 2024-10-07
Suggest people search existing issues in error message


schneems:
In our error messages, we suggest that people open an issue and do not provide a URL. Instead we should provide them with https://github.com/heroku/buildpacks-ruby/issues and suggest that people search for one. Leaving the suggestion that they open an issue implied. 

The goal is to increase the signal to noise ratio and not accidentally spam the current buildpack maintainer.

Error message is here https://github.com/heroku/buildpacks-ruby/blob/b377e25769c7287c1866cafd3e8caa4cbf70da6e/buildpacks/ruby/src/user_errors.rs#L34

Additionally we should reference or link to the application contract https://github.com/heroku/buildpacks-ruby/blob/main/docs/application_contract.md

PR #350, opened 2024-11-14
Suggest people search existing issues in error message

BrianBorge:
- Addresses https://github.com/heroku/buildpacks-ruby/issues/333 (Suggest people search existing issues in error message)


schneems (approved):
Better late than never, this addition is above and beyond. Thanks!

PR #351, opened 2024-11-16
Deny unknown fields

BrianBorge:
- There were some structs missed with the first commit to address this issue (93abd7735c8d0e8c7720eb9482126d3e3ac22659)

### DIFF
--- a/buildpacks/ruby/src/layers/bundle_install_layer.rs
+++ b/buildpacks/ruby/src/layers/bundle_install_layer.rs
@@ -162,6 +162,7 @@ impl MetadataDiff for Metadata {
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone, Eq, PartialEq)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct MetadataV1 {
     pub(crate) stack: String,
     pub(crate) ruby_version: ResolvedRubyVersion,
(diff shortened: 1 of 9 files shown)


buildpacks/ruby/src/layers/bundle_install_layer.rs
```diff
@@ -321,6 +323,7 @@ fn display_name(cmd: &mut Command, env: &Env) -> String {
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone, Eq, PartialEq, Default)]
+#[serde(deny_unknown_fields)]
```
BrianBorge, line 326: Not a metadata struct but struct implements Deserialize, so I added `#[serde(deny_unknown_fields)]`.  There were a few more examples like this.  Let me know if this needs to be reverted.

BrianBorge:
I missed these because I searched with a bad regex 😭

buildpacks/ruby/src/layers/bundle_install_layer.rs
schneems: I'm not sure that specific struct is used. Which is odd because clippy usually notices and warns on stuff like that. Having this won't block a merge.

schneems (approved):
Thanks!

PR #372, opened 2024-12-13
Add support for pre-release ruby versions

schneems:
Pre-releases in ruby use either `-` or `.` and touch the last number, versus patch identifiers start with `p`. This commit now captures identifiers such as `.rc1` or `.preview2`

The state of tooling is confusing, as different tools manipulate values differently:

```
$ cat Gemfile
source 'https://rubygems.org'
ruby "3.4.0-rc1"
gem 'rake'
⛄️ 3.4.0 🚀 /private/tmp/aa82068ba3124d84c76d7faf156888a5 (main)
$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    rake (13.2.1)

PLATFORMS
  arm64-darwin-23
  ruby

DEPENDENCIES
  rake

RUBY VERSION
   ruby 3.4.0.rc1

BUNDLED WITH
   2.5.23
```

Here's some more context:

- https://github.com/heroku/docker-heroku-ruby-builder/blob/175b5c15104cb809e1a0adc2c2a9902f94083954/lib/version_parts.rb#L1-L57
- https://github.com/heroku/docker-heroku-ruby-builder/blob/a41f31c72a74df6bcb4f9266e12192b0af83da98/shared/src/download_ruby_version.rs#L73-L93

Our versions are stored on S3 like this:

```
s3://heroku-buildpack-ruby/heroku-24/arm64/ruby-3.4.0.rc1.tgz
```

### DIFF
--- a/commons/src/gemfile_lock.rs
+++ b/commons/src/gemfile_lock.rs
@@ -116,13 +116,12 @@ impl FromStr for GemfileLock {
     type Err = std::convert::Infallible;
 
     fn from_str(string: &str) -> Result<Self, Self::Err> {
-        let bundled_with_re = Regex::new("BUNDLED WITH\\s   (\\d+\\.\\d+\\.\\d+)")
-            .expect("Internal error: Bad regex"); // Checked via clippy
+        let bundled_with_re =
+            Regex::new("BUNDLED WITH\\s   (\\d+\\.\\d+\\.\\d+)").expect("Clippy checked");
         let main_ruby_version_re =
             Regex::new("RUBY VERSION\\s   ruby (\\d+\\.\\d+\\.\\d+((-|\\.)\\S*\\d+)?)")
-                .expect("Internal error: Bad regex"); // Checked via clippy
-        let jruby_version_re =
-            Regex::new("\\(jruby ((\\d+|\\.)+)\\)").expect("Internal error: Bad regex"); // Checked via clippy
+                .expect("Clippy checked");
+        let jruby_version_re = Regex::new("\\(jruby ((\\d+|\\.)+)\\)").expect("Clippy checked");
 
         let bundler_version = match bundled_with_re.captures(string).and_then(|c| c.get(1)) {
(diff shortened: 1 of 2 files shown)


commons/src/gemfile_lock.rs
```diff
@@ -116,12 +116,12 @@ impl FromStr for GemfileLock {
     type Err = std::convert::Infallible;
 
     fn from_str(string: &str) -> Result<Self, Self::Err> {
```
edmorley, line 118: In the Slack thread, I see the error message when someone tried to use the pre-release version was an HTTP 404 rather than an error message that mentioned finding an invalid version in `Gemfile.lock`. 

Longer term, would it be worth having more explicit handling for invalid versions, so that they are rejected early with a specific "invalid version; here's the form it should take" type error, rather than the buildpack only realising its invalid via the 404?

For example, the Python CNB shows this specific error message (and in this case never tries to make an HTTP request):
https://github.com/heroku/buildpacks-python/blob/e62672366eb192750b166c2af883e10b43cbac54/tests/python_version_test.rs#L207-L222

edmorley (approved):

commons/src/gemfile_lock.rs
schneems: Currently the Ruby buildpack doesn't know what is and isn't a valid version. It takes whatever you give it and sees if it's on S3. Longer term the idea is to move to an inventory file model, but the going is slow. I'm currently able to generate inventory updates, but I need a way to bundle them all together and make a self PR. I poked at it briefly https://github.com/heroku/docker-heroku-ruby-builder/commit/204df94b3263e0340a57f08295e6b6c49d77a4e0.

Once we've got an inventory file we could levenshtein distance and say "X.Q.P" is not valid "did you mean X.Y.Z?"

The classic buildpack suggests versions by querying/walking S3, which... if we're moving to an inventory file, isn't worth implementing here if we're just going to turn around and throw it away in favor of a file.

PR #384, opened 2025-01-10
[stacked] Spike: Auto Layer Ordering

schneems:
Prefix layers with sequential numbers starting at `0001_<name>`. To support this it assumes any layers on disk with a number prefix is equivalent i.e. changing the order of layers will not invalidate the cache.

This has the benefit that:

- Layers can be named something semantic without side effects.
- Build and launch layer behavior is guaranteed to be the same (provided `read_env` is called and applied for every layer in main).


schneems:
I think it makes more sense to have a single function/macro that does it instead of trying to re-order things behind the scenes automatically, though we still need the conversion code.

schneems:
Ed pointed out that not all binaries can be relocated and might embed path information in the binary itself. I think ordering needs to be a first-class feature from CNB and opened https://github.com/buildpacks/rfcs/pull/322 to explore what the interface might look like based on my experiences here.

PR #395, opened 2025-02-11
Use global print output

schneems:
Switches from stdout to the global writer which defaults to stderr. This swich guarantees consistent print output when the print stream is dropped and then we have to construct a new one to emit an error message.

### DIFF
--- a/buildpacks/ruby/src/gem_list.rs
+++ b/buildpacks/ruby/src/gem_list.rs
@@ -5,7 +5,7 @@ use fun_run::CmdError;
 use regex::Regex;
 use std::collections::HashMap;
 use std::ffi::OsStr;
-use std::io::Stdout;
+use std::io::Write;
 use std::process::Command;
 
 /// ## Gets list of an application's dependencies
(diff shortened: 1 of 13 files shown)


colincasey (approved):
Looks good but can you explain the rationale behind switching from `stdout` to `stderr` again? What's the benefit there?

schneems:
> Looks good but can you explain the rationale behind switching from stdout to stderr again? What's the benefit there?

Sure. Here's an internal link https://salesforce-internal.slack.com/archives/C02GZCPPV38/p1738340566002289.

For everyone else, as an oversimplification: Stdout is for machines, stderr is for humans. Here's some more info https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout.

As a related sidebar: I'm wanting to propose that `$ pack` CLI re-routes all buildpack output to stderr so that we can allow using `pack build` with unix pipes by emitting only specific, structured information to stdout. There's also some ordering issues that are related:

- https://github.com/buildpacks/pack/issues/2330 (so we should default to one or the other, and stderr is the more correct output)

PR #402, opened 2025-03-11
Configure gem install location via GEM_*

schneems:
The previous logic used the `BUNDLE_PATH` environment variable to direct bundler where to install gems. This had the side effect of adding additional paths to the directory structure, so the files wouldn't be in `<layer-path>` they would be in a `<layer-path>/ruby/<major>.<minor>.0/` directory such as `<layer-path>/ruby/3.4.0`. The classic buildpack handles this by shelling out to Ruby https://github.com/heroku/heroku-buildpack-ruby/blob/b3ccc41885135ae495c604a512b523c81241914d/lib/language_pack/ruby.rb#L157. This change diverges and takes an alternative approach, using the `GEM_HOME` and `GEM_PATH` environment variables to configure installation location rather than configuring bundler directly.

When `bundle install` is run, it will install into the first `GEM_PATH` (there can be multiple). This install will be direct (without any additional directories under it). This is what we want. This also allows us to remove `BUNDLE_BIN` which is no longer needed (bundler will install binaries into `GEM_PATH/bin`). 

- `BUNDLE_DEPLOYMENT` does more than forces the `Gemfile.lock` to be frozen, it also installs gems into `vendor/bundle`, which we don't want. This has been changed to `BUNDLE_FROZEN=1`.

Unfortunately, removing `BUNDLE_PATH` causes the automatic clean after install logic (`BUNDLE_CLEAN=1`) to error with a warning saying that it is unsafe because there's no explicit path. To work around this, I added a manual call to `bundle clean --force` after the `bundle install`. This requires I also remove the `BUNDLE_CLEAN=1` environment variable.

Because this changes the structure of the underlying gems on disk, I need to clear the cache manually by updating the key to `v2`.

### DIFF
--- a/buildpacks/ruby/CHANGELOG.md
+++ b/buildpacks/ruby/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Gem install behavior and configuration (https://github.com/heroku/buildpacks-ruby/pull/402)
+  - Gem install path is now configured with `GEM_HOME` and `GEM_PATH` instead of `BUNDLE_PATH`.
+  - Cleaning gems is now accomplished via running `bundle clean --force`. Previously it was accomplished by setting `BUNDLE_CLEAN=1`.
+  - The `BUNDLE_DEPLOYMENT=1` environment variable is changed to `BUNDLE_FROZEN=1`.
+  - The `BUNDLE_BIN` environment variable is no longer set.
+
 ## [5.1.0] - 2025-02-28
 
 ### Changed
--- a/buildpacks/ruby/CHANGELOG.md
+++ b/buildpacks/ruby/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Gem install behavior and configuration (https://github.com/heroku/buildpacks-ruby/pull/402)
+- Gem install behavior and configuration ([#402](https://github.com/heroku/buildpacks-ruby/pull/402))
   - Gem install path is now configured with `GEM_HOME` and `GEM_PATH` instead of `BUNDLE_PATH`.
   - Cleaning gems is now accomplished via running `bundle clean --force`. Previously it was accomplished by setting `BUNDLE_CLEAN=1`.
   - The `BUNDLE_DEPLOYMENT=1` environment variable is changed to `BUNDLE_FROZEN=1`.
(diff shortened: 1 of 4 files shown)


buildpacks/ruby/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Gem install behavior and configuration (https://github.com/heroku/buildpacks-ruby/pull/402)
```
edmorley, line 12: (for consistency with the earlier entries)

```suggestion
- Gem install behavior and configuration ([#402](https://github.com/heroku/buildpacks-ruby/pull/402))
```

buildpacks/ruby/src/layers/bundle_install_layer.rs
```diff
@@ -306,14 +312,7 @@ fn display_name(cmd: &mut Command, env: &Env) -> String {
...
         env,
-        [
-            "BUNDLE_BIN",
-            "BUNDLE_CLEAN",
-            "BUNDLE_DEPLOYMENT",
-            "BUNDLE_GEMFILE",
-            "BUNDLE_PATH",
-            "BUNDLE_WITHOUT",
-        ],
+        ["BUNDLE_FROZEN", "BUNDLE_GEMFILE", "BUNDLE_WITHOUT"],
```
edmorley, line 315: (optional) Reading the comment for `GEM_HOME` on line 273, it seems Bundler will listen to the value for `GEM_HOME` - so should `GEM_HOME` now be included in the list here too in lieu of `BUNDLE_PATH`?

edmorley (approved):

buildpacks/ruby/src/layers/bundle_install_layer.rs
schneems: Short: It's a good question. The answer today is: we shouldn't include it.

Full:
I considered it, but I think that would be confusing to most Ruby developers. If anything, this PR is a move closer to how people have their systems configured locally, for example, chruby is the version switcher I use https://github.com/postmodern/chruby/blob/a543a35790e5528b5a67de20e78a7390f5f7606e/share/chruby/chruby.sh#L55-L56. 

One issue I ran into that you can see in the (privately linked and communicated slack thread) is that the customer tried copying and pasting the bundle install command wholesale without changing any paths. Having GEM_PATH or GEM_HOME in the output would cause people to run a command that won't work locally or may work, but then not behave as expected for future installs. 

As a semi-related side note, setting GEM_HOME doesn't seem to do anything by itself if it's not on the GEM_PATH as well. These configuration options are woefully underdocumented, and I've mostly had to figure out how they work by looking at the source for existing tooling and experimentation. So if we advertise one, I think we need to advertise both.

I think that today there's little value in announcing how we're configuring things, only that we are. The only somewhat unusual thing that I can think that we're doing today is putting bundler in a different location than GEM_HOME.

PR #412, opened 2025-03-24
Initial honeycomb tracing

schneems:
Add tracing

Related: 

- https://github.com/heroku/buildpacks-go/pull/348
- https://github.com/heroku/buildpacks-deb-packages/pull/104

GUS-W-18112178

### DIFF
--- a/buildpacks/ruby/Cargo.toml
+++ b/buildpacks/ruby/Cargo.toml
@@ -37,6 +37,7 @@ ureq = { version = "2", default-features = false, features = ["tls"] }
 url = "2"
 magic_migrate = "1.0"
 toml = "0.8"
+tracing = "0.1.41"
 
 [dev-dependencies]
 libcnb-test = "=0.28.1"
--- a/buildpacks/ruby/Cargo.toml
+++ b/buildpacks/ruby/Cargo.toml
@@ -37,7 +37,7 @@ ureq = { version = "2", default-features = false, features = ["tls"] }
 url = "2"
 magic_migrate = "1.0"
 toml = "0.8"
-tracing = "0.1.41"
+tracing = "0.1"
 
 [dev-dependencies]
 libcnb-test = "=0.28.1"
--- a/buildpacks/ruby/src/layers/bundle_download_layer.rs
+++ b/buildpacks/ruby/src/layers/bundle_download_layer.rs
@@ -83,7 +83,7 @@ pub(crate) struct MetadataV1 {
     pub(crate) version: ResolvedBundlerVersion,
 }
 
-#[tracing::instrument(skip(bullet, env, gem_path), err)]
+#[tracing::instrument(skip_all)]
 fn download_bundler<W>(
     mut bullet: Print<SubBullet<W>>,
     env: &Env,
--- a/buildpacks/ruby/src/layers/bundle_download_layer.rs
+++ b/buildpacks/ruby/src/layers/bundle_download_layer.rs
@@ -66,7 +66,8 @@ where
                     bullet = bullet.sub_bullet(cause);
                 }
             }
-            bullet = download_bundler(bullet, env, metadata, &layer_ref.path())?;
+            bullet = download_bundler(bullet, env, &metadata.version, &layer_ref.path())
+                .map_err(RubyBuildpackError::GemInstallBundlerCommandError)?;
         }
     }
     Ok((bullet, layer_ref.read_env()?))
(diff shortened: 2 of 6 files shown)


buildpacks/ruby/Cargo.toml
```diff
@@ -31,6 +31,7 @@ ureq = { version = "2", default-features = false, features = ["tls"] }
 url = "2"
 magic_migrate = "1.0"
 toml = "0.8"
+tracing = "0.1.41"
```
edmorley, line 34:
```suggestion
tracing = "0.1"
```

buildpacks/ruby/src/layers/bundle_download_layer.rs
```diff
@@ -82,20 +83,21 @@ pub(crate) struct MetadataV1 {
     pub(crate) version: ResolvedBundlerVersion,
 }
 
+#[tracing::instrument(name = "download_bundler", skip(bullet, env, gem_path), err)]
```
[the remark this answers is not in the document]
colincasey: Just stick to the [semantic conventions for attribute naming](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md) and [span guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span) from Open Telemetry. 

Though we may want to use a namespacing convention for our buildpacks like `cnb.{lang}.{attribute}`. E.g.; `cnb.ruby.bundler.version`.

@joshwlewis any other suggestions you'd make here?

buildpacks/ruby/src/main.rs
```diff
@@ -126,7 +126,9 @@ impl Buildpack for RubyBuildpack {
             .map_err(|error| RubyBuildpackError::MissingGemfileLock(lockfile, error))?;
         let gemfile_lock = GemfileLock::from_str(&lockfile_contents).expect("Infallible");
         let bundler_version = gemfile_lock.resolve_bundler("2.5.23");
+        tracing::info!(version_bundler = bundler_version.to_string());
```
colincasey, line 129: As this is a macro, it doesn't need to obey Rust naming rules which are slightly different than OTel naming standards. 

```suggestion
        tracing::info!(cnb.ruby.bundler.version = bundler_version.to_string());
```

buildpacks/ruby/src/main.rs
```diff
@@ -126,7 +126,9 @@ impl Buildpack for RubyBuildpack {
...
         let ruby_version = gemfile_lock.resolve_ruby("3.3.7");
+        tracing::info!(version_ruby = ruby_version.to_string());
```
colincasey, line 131: Similar to the other comment:

```suggestion
        tracing::info!(cnb.ruby.runtime.version = ruby_version.to_string());
```

edmorley:
This was marked as skip-changelog, however, I think we should still have a changelog entry for it, since:
- it does have an effect on the end build (e.g. it's not just a refactor or testcase change)
- it will help us work out which buildpacks have added tracing and when - which will help when we e.g. can't find traces in Honeycomb or wonder when various attributes were added or removed (I've often found myself looking at the classic Python buildpack changelog when interpreting Honeycomb results)
- seeing changelog entries under "unreleased" acts as a reminder that there are things on `main` that need releasing (otherwise I think it's easy to presume the only changes are not things that need to be released, such as adding tests or refactors)
- tracing/telemetry is an area where I think perceived transparency is important (even though the telemetry is privacy preserving and the whole codebase and PRs are public already)

edmorley:
> This was marked as skip-changelog, however, I think we should still have a changelog entry for it, since:

I think a rough rule of thumb we could use across all of our buildpack repos is...

"Would you release a new buildpack version even if this change was the only thing on `main`?"

If the answer is no (eg because the PR just adds a testcase or updates the README), the  `skip-changelog` is likely appropriate.

If the answer is yes, then the change should probably have a changelog entry - because (a) the change clearly has an effect on the published buildpack if it needs releasing, (b) it would be confusing to end users if we publish a new version that has "No changes" listed in the changelog - so IMO having an entry there makes sense even for ~internal ish changes (let alone those that have a more visible effect for end users) :-)

buildpacks/ruby/src/layers/bundle_download_layer.rs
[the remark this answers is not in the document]
schneems: Looking at the deb-packages buildpack it doesn't specify a name (https://github.com/heroku/buildpacks-deb-packages/pull/104/files). The docs from `instrument` macro suggest it then uses the name of the function. If I remove my `name = download_bundler` here then I think the result is the same.

buildpacks/ruby/src/layers/bundle_download_layer.rs
```diff
@@ -82,20 +83,21 @@ pub(crate) struct MetadataV1 {
     pub(crate) version: ResolvedBundlerVersion,
 }
 
+#[tracing::instrument(skip(bullet, env, gem_path), err)]
```
colincasey, line 86: I'm curious what the result of including `err` in the trace here will provide. I mean, apart from having the error attached directly to this span. Because of how error handling in the CNBs works, you may want to consider centralizing the capturing of errors to the `on_error` handler. 

I don't consider this a problem that needs to be fixed, it's just a different approach than I was expecting to see.

colincasey, line 86: Namespacing the `version` and `gem_path` would be my only suggestion here but that's something that, as a team, I think needs more discussion so feel free to proceed without that for now.

colincasey (approved):

buildpacks/ruby/src/layers/bundle_download_layer.rs
schneems: I copied this from the go buildpack originally. I'll just disable all of them.

PR #425, opened 2025-05-16
Run `bundle install` on every push

schneems:
Previously: We inspected the state of the last `Gemfile.lock` and user environment variables to only run `bundle install` when something changed. This provides minimal value in practice and requires a lot of plumbing. Now we call `bundle install` on every push. This adds a (relatively) negligible overhead (0.3~1s) per deploy for a fully cached app, and allows us to remove a lot of code.

- Deprecates `MetadataDigest` that was introduced to support this feature
- Removes old Metadata structs introduced before Fir went GA (details in commits)
- Removes logic used to support old metadata

Changes are structured in ordered commits with explanatory messages

GUS-W-18540827

### DIFF
--- a/buildpacks/ruby/src/layers/bundle_install_layer.rs
+++ b/buildpacks/ruby/src/layers/bundle_install_layer.rs
@@ -272,16 +272,6 @@ mod test {
     /// When the vec is empty the cache is kept, otherwise it is invalidated
     #[test]
     fn metadata_diff_messages() {
-        let tmpdir = tempfile::tempdir().unwrap();
-        let app_path = tmpdir.path().to_path_buf();
-        let gemfile = app_path.join("Gemfile");
-        let env = Env::new();
-        let context = FakeContext {
-            platform: FakePlatform { env },
-            app_path,
-        };
-        std::fs::write(&gemfile, "iamagemfile").unwrap();
-
         let old = Metadata {
             ruby_version: ResolvedRubyVersion("3.5.3".to_string()),
             os_distribution: OsDistribution {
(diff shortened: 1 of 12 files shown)


buildpacks/ruby/src/layers/bundle_install_layer.rs
```diff
@@ -53,211 +36,103 @@ pub(crate) fn call(
...
-    ///
-    /// This value is cached with metadata, so changing the struct
-    /// may cause metadata to be invalidated (and the cache cleared).
-    ///
-    #[cache_diff(ignore)]
-    pub(crate) digest: MetadataDigest, // Must be last for serde to be happy https://github.com/toml-rs/toml-rs/issues/142
 }
 
-fn clear_v1(_new: &Metadata, old: &Metadata) -> Vec<String> {
+fn clear_v1(_new: &MetadataV3, old: &MetadataV3) -> Vec<String> {
```
colincasey, line 115: Very minor - I found this method name slightly confusing in the context of multiple `MetadataV*` structs + migrations + diffing. My first thought was that this method maybe clears `MetadataV1` stuff from `MetadataV3` as part of a migration process but that's obviously not the case given the argument types. I had to jump through the code a bit to realize it's for the diff output. Maybe this function could be called something more descriptive?

buildpacks/ruby/src/layers/bundle_install_layer.rs
```diff
@@ -53,211 +36,103 @@ pub(crate) fn call(
...
+            .env_clear()
+            .envs(&env),
+    )
+    .map_err(RubyBuildpackError::BundleInstallCommandError)?;
 
     layer_ref.read_env()
 }
 
-pub(crate) type Metadata = MetadataV3;
+pub(crate) type Metadata = MetadataV4;
```
colincasey, line 78: Small nitpick - I think this type alias could be named `CurrentMetadataVersion`,  `LatestMetadata`, or something to that affect. Only because my first instinct when reading some of the function signatures was that `Metadata` might be the first struct you started with (i.e.; V0) before realizing you'd need subsequent versions.

colincasey (approved):

buildpacks/ruby/src/layers/bundle_install_layer.rs
schneems: I'm not sure how much it helps, but I moved the type alias to the top of the file and added a comment on it. I'm hesitant to apply an adjective here.

buildpacks/ruby/src/layers/bundle_install_layer.rs
schneems: TBF I got confused with it too for the same reason in refactoring. So I agree. I updated in https://github.com/heroku/buildpacks-ruby/pull/425/commits/7025309dde0080b61444941ee8a6fcef17422047. This names things more in line with their behaviors than their intent. In theory, additional logic could go here without needing to rename the function, which is a property I like.

The only downside is if we needed different custom cache clear functions for different structs, then this is the "best" name so we would have to come up with some convention, but matching it to the struct version seems good.

